PR #4, opened 2019-12-03
fixed additionalProperties

WaleedAshraf:


WaleedAshraf:
I did that intentionally so that we can release a patch version. I thought that what you meant on slack. Maybe I understood it wrong. ✌️ 👍 

> Feel free to patch the asyncapi package but let's not do any repo separation now please.

fmvilas:
Yeah, I'm actually releasing it but there was a need to update package-lock.json too and I had to tag the version for deployment. That's why I'm saying. Not that it's a problem 👍

WaleedAshraf:
I was thinking should we pull schema from asyncapi repo here? Instead of copy/paste.
That is also related if/when we move schem.json out of asyncapi repo. So, there's a lot to do.

fmvilas:
Yes, that might be a better solution. Having a single source of truth would help. We'd just need to have a "compilation" script that will download the file from the AsyncAPI repo. We could even automate creating a PR in this repo every time the file changes there :)

WaleedAshraf:
other thing we can do is we don't need this repo at all. It just has the schemas as in asyncapi repo. So why not publish that repo as a package?

If we put these lines in package.json of that repo, it'll only include schemas in npm package, nothing else. 
```
"files": [
    "versions/1.0.0/schema.json",
    "versions/1.1.0/schema.json",
    "versions/1.2.0/schema.json",
    "versions/2.0.0/schema.json"
  ],
```
I understand, that that repo is for schema, not a node package, but we already have package.json and dependencies and tests in that repo.

fmvilas:
Because I don't want to bloat the other repo and because schemas will be needed in different programming languages, not just Node.js. We can probably remove the "node" from this repo and convert it into a Nodejs package, a Maven and Gradle package (or whatever it's called), a PHP Composer package, Go package, etc.

PR #6, opened 2020-03-20
feat: add configuration for automated releases

derberg:
Same as https://github.com/asyncapi/generator/pull/242 except of:
- no docker release
- no contribution guide and pr template as this will be handled in https://github.com/asyncapi/asyncapi/issues/355
- added step to run `npm test`
- avoided mistakes done with generator - package marked as public and removed travis and updated readme

See also: https://github.com/asyncapi/parser-js/issues/52

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "description": "AsyncAPI schema versions",
   "main": "index.js",
   "scripts": {
-    "test": "nyc mocha"
+    "test": "nyc mocha",
+    "release": "semantic-release"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -1,10 +1,11 @@
 {
-  "name": "asyncapi",
+  "name": "@asyncapi/asyncapi",
```
fmvilas: What about `@asyncapi/versions` instead?

derberg: I don't really have an opinion here, version sounds strange though, how about `spec` or `spec-schema` ?

fmvilas: `@asyncapi/specs` maybe?

PR #21, opened 2020-10-09
chore: Cleaned up automerge workflow for dependabot

C-Zekeri:
**Description**
- Updated actors for autoapprove

**Related issue(s)**
See https://github.com/asyncapi/asyncapi-node/pull/20#issuecomment-705646435


derberg (approved):
Thanks so much!

PR #47, opened 2021-03-17
fix: allow extending components object with specification extensions

### LINKED ISSUE #46: Schema for v2.0.0 does not allow to extend the components object with specification extensions
#### Describe the bug
The schema for v2.0.0 (and all older versions) does not allow to extend the [components object](https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#L251-L332) with specification extensions. This is inconsistent with the [specification](https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#componentsObject) which states that the components object can be extended with specification extensions. 

#### How to Reproduce
Validating the following exemplary AsyncAPI document against the [v2.0.0 schema](https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json) with any JSON schema validation library results in an error like _Property 'x-extension' has not been defined and the schema does not allow additional properties_ or similar:
```json
{
    "asyncapi": "2.0.0",
    "info": {"version": "1.0", "title": "Example API"},
    "channels": {},
    "components": {
        "x-extension": "xxx"
    }
}
```

#### Expected behavior
Above AsyncAPI document should be validated without errors.

balogal:
**Description**
According to the specification the components object can be extended with specification extensions. This change realizes the specified behavior

**Related issue(s)**
Fixes #46

### DIFF
--- a/schemas/2.0.0.json
+++ b/schemas/2.0.0.json
@@ -252,6 +252,11 @@
       "type": "object",
       "description": "An object to hold a set of reusable objects for different aspects of the AsyncAPI Specification.",
       "additionalProperties": false,
+      "patternProperties": {
+        "^x-[\\w\\d\\.\\-\\_]+$": {
+          "$ref": "#/definitions/specificationExtension"
+        }
+      },
       "properties": {
         "schemas": {
           "$ref": "#/definitions/schemas"


derberg (approved):

derberg:
@balogal thanks a lot for the fix.

Also, keep in mind that we have a place where we want to keep all the extensions so they can be shared by the community and once it is adopted, it is a candidate to get merged with the spec. Long term we want to support those extensions with the JS parser. This is the place https://github.com/asyncapi/extensions-catalog

PR #54, opened 2021-05-03
feat: add name and summary properties to message examples

lbroudoux:
**Description**

Complete message examples with `name` and `summary` attributes in order to:

* provide better documentation for examples,
* allow development of tooling leveraging examples (see Microcks.io)

This is implementing the [latest proposition](https://github.com/asyncapi/spec/issues/329#issuecomment-823993976) we had when discussing the topic. It does not introduce breaking changes and thus can be applied to current `2.0.0`.

**Related issue(s)**

This contributes to asyncapi/spec#329

### DIFF
--- a/schemas/2.0.0.json
+++ b/schemas/2.0.0.json
@@ -703,6 +703,14 @@
                     "type": "object",
                     "additionalProperties": false,
                     "properties": {
+                      "name": {
+                        "type": "string",
+                        "description": "Name of the message example."
+                      },
+                      "summary": {
+                        "type": "string",
+                        "description": "A brief summary of the message example."
+                      },
                       "headers": {
                         "type": "object"
                       },
--- a/schemas/2.0.0.json
+++ b/schemas/2.0.0.json
@@ -702,14 +702,10 @@
                   "items": {
                     "type": "object",
                     "additionalProperties": false,
-                    "dependencies": {
-                      "name": {
-                        "required": ["payload"]
-                      },
-                      "summary": {
-                        "required": ["payload"]
-                      }
-                    },
+                    "anyOf": [
+                      {"required": ["payload"] },
+                      {"required": ["headers"] } 
+                    ],
                     "properties": {
                       "name": {
                         "type": "string",
--- a/schemas/2.0.0.json
+++ b/schemas/2.0.0.json
@@ -713,7 +713,7 @@
                     "properties": {
                       "name": {
                         "type": "string",
-                        "description": "Machine reeadable name of the message example."
+                        "description": "Machine readable name of the message example."
                       },
                       "summary": {
                         "type": "string",
--- a/schemas/2.0.0.json
+++ b/schemas/2.0.0.json
@@ -702,10 +702,18 @@
                   "items": {
                     "type": "object",
                     "additionalProperties": false,
+                    "dependencies": {
+                      "name": {
+                        "required": ["payload"]
+                      },
+                      "summary": {
+                        "required": ["payload"]
+                      }
+                    },
                     "properties": {
                       "name": {
                         "type": "string",
-                        "description": "Name of the message example."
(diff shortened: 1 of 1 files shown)


schemas/2.0.0.json
```diff
@@ -703,6 +703,14 @@
                     "type": "object",
                     "additionalProperties": false,
                     "properties": {
+                      "name": {
+                        "type": "string",
+                        "description": "Name of the message example."
```
derberg, line 708: can you please specify it should be machine readable name

schemas/2.0.0.json
```diff
@@ -703,6 +703,14 @@
                     "type": "object",
                     "additionalProperties": false,
                     "properties": {
+                      "name": {
```
derberg, line 706: looking at the current spec description, in my opinion because we have new fields here, we need schema that validates if only `name` and/or `summary` is added, that `headers` or `payload` are required.

What do you think?

schemas/2.0.0.json
lbroudoux: Sure. Pushing an update.

schemas/2.0.0.json
lbroudoux: Good idea. Adding `dependencies` on both `name` and `summary` so that `payload` is then required. I would say that `headers` remain optional in any cases. Pushing an update.

schemas/2.0.0.json
```diff
@@ -702,10 +702,18 @@
...
+                      },
+                      "summary": {
+                        "required": ["payload"]
+                      }
+                    },
                     "properties": {
                       "name": {
                         "type": "string",
-                        "description": "Name of the message example."
+                        "description": "Machine reeadable name of the message example."
```
MikeRalphson, line 716: Typo -> `readable`

lbroudoux: Oops... Now fixed.

MikeRalphson (approved):

schemas/2.0.0.json
derberg: but now, doesn't it mean that if users just want to provide an example of `headers` only and `name`, they are forced to provide `payload` example too?

lbroudoux: Yes that's what it mean, you're right. Is there any cases we'd like to give named example without any payload? This may lead to inconsistent examples imho

derberg: named example without any example, no, doesn't make sense, but it makes sense that someone just wants to provide an example of the headers only, and this is what is written in the spec in your proposal, so you would need to adjust the `dependency` keyword, that it is either `payload` or `headers` that must be provided when `name` or `summary` are there

lbroudoux: Seems legit. However I think I stuck to just putting `payload` as required as I don't know if there's a way to say it's either `payload` or `headers`... Whatever the construction used (dependencies, anyOf, if-then-else) I cannot figure out a way to do that in JSON schema ...

derberg: @lbroudoux I asked on JSON Schema Slack, maybe the will have an idea

derberg: @magicmatatjahu proposed something like 
```
items: {
  type: object
  properties: {
    ...payload, name, summary, headers
  },
  anyOf: [
    { required: ['payload'] }
    { required: ['headers'] }
  ]
}
```

you would probably have to test it 😄

lbroudoux: It seems to work well! Expect an update on this PR in few minutes.

derberg (approved):
LGTM 🚀 
thanks @magicmatatjahu for helping out!

magicmatatjahu (approved):
🚀

fmvilas (approved):
🚀

PR #59, opened 2021-05-19
chore: enable releases for spec-related release branches

derberg:
see also https://github.com/asyncapi/spec/issues/536

it is supposed to work the same as here https://github.com/asyncapi/spec/pull/539 + npm release

relevant workflow configuration is in master already as it came from a global workflow https://github.com/asyncapi/asyncapi-node/commit/5c007ed02d2c625b001f6daa4a7827117a28a5ca

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,15 +38,19 @@
     "mocha": "^8.2.1",
     "nyc": "^15.1.0",
     "@semantic-release/commit-analyzer": "^8.0.1",
-    "@semantic-release/github": "^7.0.4",
+    "@semantic-release/github": "7.2.3",
     "@semantic-release/npm": "^7.0.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -38,15 +38,19 @@
     "mocha": "^8.2.1",
     "nyc": "^15.1.0",
     "@semantic-release/commit-analyzer": "^8.0.1",
-    "@semantic-release/github": "^7.0.4",
+    "@semantic-release/github": "7.2.3",
     "@semantic-release/npm": "^7.0.3",
     "@semantic-release/release-notes-generator": "^9.0.1",
```
magicmatatjahu, line 43: You can update also this package to `9.0.2`, but it's optional.

magicmatatjahu (approved):

PR #63, opened 2021-06-02
fix: support for Boolean JSON Schema

magicmatatjahu:
**Description**

As in title. As we support JSON Schema draft-07 as schema, then we should also support `true/false` schemas in JSON Schema of our spec. More info https://github.com/asyncapi/parser-js/issues/232

I removed `"type": "object",` from schema, because we should infer `type` field from `"http://json-schema.org/draft-07/schema#"`. Currently we override it.

**Related issue(s)**
See more https://github.com/asyncapi/parser-js/issues/232
Blocked by https://github.com/asyncapi/spec/pull/550

### DIFF
--- a/schemas/2.0.0.json
+++ b/schemas/2.0.0.json
@@ -362,7 +362,6 @@
           "$ref": "http://json-schema.org/draft-07/schema#"
         },
         {
-          "type": "object",
           "patternProperties": {
             "^x-[\\w\\d\\.\\-\\_]+$": {
               "$ref": "#/definitions/specificationExtension"


schemas/2.0.0.json
```diff
@@ -362,7 +362,6 @@
           "$ref": "http://json-schema.org/draft-07/schema#"
         },
         {
-          "type": "object",
```
fmvilas, line 365: I'd not remove this line but instead, I'd do the following:

```json
"schema": { // Current line 359
  "oneOf": [
    { "type": "boolean" },
    { "allOf": [ // Current line 360
        ...
```

fmvilas (changes requested):

schemas/2.0.0.json
magicmatatjahu: It's one the solution, but when we will remove the `type` from second schema, then we infer the type from `http://json-schema.org/draft-07/schema` and it is:

```json
"type": ["object", "boolean"],
```

magicmatatjahu: I think that `oneOf` can only complicated (already complex) our JSON Schema

fmvilas: You're right. Forgot that `type` can take an array. Go for it then.

fmvilas: Fore sure not removing the `type: object` line because that would mean we accept anything and that's not true.

magicmatatjahu: But the `"type": ["object", "boolean"]` is inferred from the `http://json-schema.org/draft-07/schema` `allOf` combined schema means that for every sub schema input data must be valid, so if I removed the `type` from the second schema then I have allowed types for first schema `object` and `boolean and for second I have allowed of course everything, but the, as I wrote, the http://json-schema.org/draft-07/schema schema only allows the [object, boolean], so at the end if someone put as input the number type it will be invalid:
- first schema - invalid
- second schema - valid
If someone put boolean or type it will be valid always :)

You can test it with this simple schema:

```
{
  allOf: [
    {
      type: ['object', 'boolean'] 
    },
    {}
  ],
}
```

in the https://www.jsonschemavalidator.net/

magicmatatjahu: But for you I will write for second schema the `object, boolean` type 😘

fmvilas: Ohh sorry, I didn't notice. Nevermind, it's fine as it is 👍

fmvilas (approved):

magicmatatjahu:
@fmvilas Thanks!

PR #75, opened 2021-06-29
feat: use release 2.1.0 of the spec

derberg:
https://github.com/asyncapi/spec/issues/536

### DIFF
--- a/schemas/2.1.0.json
+++ b/schemas/2.1.0.json
@@ -702,7 +702,19 @@
                   "items": {
                     "type": "object",
                     "additionalProperties": false,
+                    "anyOf": [
+                      {"required": ["payload"] },
+                      {"required": ["headers"] } 
+                    ],
                     "properties": {
+                      "name": {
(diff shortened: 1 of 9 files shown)


schemas/2.1.0.json
```diff
@@ -0,0 +1,1477 @@
...
+      },
+      "description": "JSON objects describing re-usable channel parameters."
+    },
+    "schema": {
+      "allOf": [
+        {
+          "$ref": "http://json-schema.org/draft-07/schema#"
+        },
+        {
+          "type": "object",
```
magicmatatjahu, line 365: I removed it in https://github.com/asyncapi/asyncapi-node/pull/63 Do we should fix it now or rather after release and make `fix:`?

derberg: no, should be fixed here too, can you open a PR and I will merge

magicmatatjahu (approved):

PR #85, opened 2021-08-11
fix: fix Message Trait examples field

char0n:
Refs https://github.com/asyncapi/spec/issues/604
Refs https://github.com/asyncapi/spec/pull/608


fmvilas (approved):

fmvilas:
@derberg @magicmatatjahu mind having a look too?

magicmatatjahu (approved):

derberg (approved):
I only removed `schema` scope from fix prefix as the whole repo is about schema anyway, and tbh we did not use scopes for `fix` and `feat` and I'm not sure what effect it has on release process 😅

PR #95, opened 2021-09-16
chore: update release branch with latest master

derberg:
updating release branch to have there needed configuration for prelease creation

### DIFF
--- a/.github/workflows/automerge-orphans.yml
+++ b/.github/workflows/automerge-orphans.yml
@@ -38,7 +38,13 @@ jobs:
             };
             const { repository: { pullRequests: { nodes } } } = await github.graphql(query, variables);
 
-            let orphans = nodes.filter((pr)=> pr.state === 'OPEN' && (pr.author.resourcePath === '/asyncapi-bot' || pr.author.resourcePath === '/apps/dependabot'))
+            let orphans = nodes.filter((pr)=> {
+              console.log('PR', pr.url);
+              console.log('State', pr.state);
+              console.log('Author resource path', pr.author.resourcePath);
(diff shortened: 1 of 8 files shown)


.github/workflows/automerge-orphans.yml
```diff
@@ -20,14 +20,13 @@ jobs:
           script: |
             const query = `query($owner:String!, $name:String!) {
               repository(owner:$owner, name:$name){
-                pullRequests(first: 100){
+                pullRequests(first: 100, states: OPEN){
```
magicmatatjahu, line 23: As I see in docs, `states` is array:

image

So probably it should be:

```
pullRequests(first: 100, states: [OPEN]){
```

derberg: you know it is just an update from master, it is already on master, replicated from .github repo, used for at least 2 weeks, works fine so far https://github.com/asyncapi/generator/runs/3616407238?check_suite_focus=true and anyway I remember I tested in in [GitHub GraphQL Explorer](https://docs.github.com/en/graphql/overview/explorer)

magicmatatjahu: If it works, then ok. I didn't do a previous review, so it looks strange for me that field points to array values, and you use single.

magicmatatjahu (approved):

.github/workflows/automerge-orphans.yml
derberg: explorer did not complain 😄

magicmatatjahu: `U mnie działa` 😏

PR #119, opened 2021-11-02
fix: revert use HTTPS on $schema urls to avoid mixed content

smoya:
Reverts asyncapi/spec-json-schemas#115

Tools like ajv (schema validator) are not compatible with https. See https://github.com/ajv-validator/ajv/issues/1104.

Also from a point of view of JSON Schema, this note about Schema ID's is very important:
> Even though schemas are identified by URIs, those identifiers are not necessarily network-addressable. They are just identifiers. Generally, implementations don’t make HTTP requests (https://) or read from the file system (file://) to fetch schemas. Instead, they provide a way to load schemas into an internal schema database. When a schema is referenced by it’s URI identifier, the schema is retrieved from the internal schema database.

Meaning it **should not** matter what protocol the URI has because It should not be treated as network address. However, it seems the [Monaco yaml plugin](https://github.com/remcohaszing/monaco-yaml) used in [Studio](https://github.com/asyncapi/studio) fills a request to the schema when the schema is not loaded from start (when it is not provided from the studio).

@magicmatatjahu is working on the fix.

### DIFF
--- a/schemas/1.0.0.json
+++ b/schemas/1.0.0.json
@@ -1,7 +1,7 @@
 {
   "title": "AsyncAPI 1.0 schema.",
   "id": "http://asyncapi.hitchhq.com/v1/schema.json#",
-  "$schema": "https://json-schema.org/draft-04/schema#",
+  "$schema": "http://json-schema.org/draft-04/schema#",
   "type": "object",
   "required": [
     "asyncapi",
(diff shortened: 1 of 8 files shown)


magicmatatjahu:
PR on Studio https://github.com/asyncapi/studio/pull/159

magicmatatjahu (approved):

PR #126, opened 2021-11-03
fix: add v2 suffix to Go module

smoya:
**Description**

This change will make the Go module compatible with Go semantic versioning.
As the tag it is >= v2, Go requires `v2` suffix in module name. See https://golang.org/ref/mod#major-version-suffixes

I also took this opportunity to add few lines in the `README.md` explaining how to install and use the Go package.

**Related issue(s)**
The issue is that users won't be able to import the module ever due to the following error:
> invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2


jonaslagoni:
What do you need to progress here @smoya?

smoya:
THis is blocked by https://github.com/asyncapi/spec-json-schemas/issues/130

derberg (approved):

Issue #120, opened 2021-11-02
Provide JSON Schema draft 7 file as part of schemas


smoya:
#### Reason/Context
[JSON Schema Draft 7](https://json-schema.org/draft-07/json-schema-release-notes.html) is the JSON Schema version AsyncAPI is compatible with. 
All our schemas are written with such dialect, ergo we specify it through the [$schema](https://json-schema.org/understanding-json-schema/reference/schema.html#id4) field.
We also make references to that same schema when referencing through `$ref`.

As you can notice, we strongly depend on such JSON Schema schema. 
I'm suggesting we include it as part of our schemas collection so any tool can directly load all schemas, including the JSON Schema Draft 7 on their Schema registries so they don't need to fetch it externally.

A recent real use case can be found https://github.com/asyncapi/studio/pull/159, where the JSON Schema Draft 7 should be provided to Monaco editor YAML plugin in order to avoid fetching it through a HTTP Request.

The plan will also include one JSON Schema schema file per version we require, meaning if we ever adopt a newer JSON Schema draft, we will include it as well here.

#### Description
Add a new file into [/schemas](https://github.com/asyncapi/spec-json-schemas/tree/master/schemas) called `json-schema-draft-07.json` with the exact content of http://json-schema.org/draft-07/schema.

jonaslagoni:
Because they are a remote reference they are bundled automatically in the bundled schema https://github.com/jonaslagoni/spec-json-schemas/blob/dd457977ac91599ad6833599f2ea6dc4c92b5a13/schemas/2.3.0.json#L482

smoya:
> Because they are a remote reference they are bundled automatically in the bundled schema https://github.com/jonaslagoni/spec-json-schemas/blob/dd457977ac91599ad6833599f2ea6dc4c92b5a13/schemas/2.3.0.json#L482

Oh I see now! Yeah, then this is gonna be closed by https://github.com/asyncapi/spec-json-schemas/pull/128.
Not sure if you want to close this one or rather add a "Fixes https://github.com/asyncapi/spec-json-schemas/issues/120" statement on your PR (I think it's better).

jonaslagoni:
> Not sure if you want to close this one or rather add a "Fixes #120" statement on your PR (I think it's better).

Already there 😄

PR #128, opened 2021-11-11
feat!: split out definitions

### LINKED ISSUE #127: Split out definitions into separate files
#### Reason/Context
To make it more digestible and easier to pinpoint problems, I think it would make sense to split each [definition](https://github.com/asyncapi/spec-json-schemas/blob/6ca4574270bce02d9e9470454f842be5e917e4c3/schemas/2.2.0.json#L58) into separate files.

#### Description
That said, I also see benefits in bundling all these separate definitions into one file. This way it remains easy to use in other settings, where you don't need all the separate files.

jonaslagoni:
## NOTICE: THIS WILL TRIGGER A MAJOR VERSION CHANGE

**Description**
This PR helps ease the maintainability of the JSON Schema document for the specification since we now have small sub-schemas where the bundling process is now automated.

Changes introduced:
- All AsyncAPI versions are now [fully separated into small sub-schemas](https://github.com/jonaslagoni/spec-json-schemas/tree/feature/improve_maintainability/definitions).
- The small sub-schemas are now accurately bundled together to include **ALL** referenced schemas through the [bundler tool](https://github.com/jonaslagoni/spec-json-schemas/tree/feature/improve_maintainability/tools/bundler). Fixes https://github.com/asyncapi/spec-json-schemas/issues/120.
- The `generate:assets` script that is run on each release triggers the generation of the bundled files, ensuring they are always up to date.
- Provides a simple script `startNewVersion` to [create new version](https://github.com/jonaslagoni/spec-json-schemas/tree/feature/improve_maintainability#creating-a-new-version).
- We are changing all id's (`$id` and `id` properties) to follow `http://asyncapi.com/definitions/<version>/<name>.json` instead of `<name>` while providing a [migration guide](https://github.com/jonaslagoni/spec-json-schemas/blob/feature/improve_maintainability/migrations/Migrate%20to%20version%203.md). The reason for this is to follow an accurate bundling behavior from JSON Schema and made the most sense.

Bugs it solves:
- Specification extension pattern property regex needs to be valid against `u` flag. This means changing `^x-[\\w\\d\\.\\-\\_]+$` to `^x-[\\w\\d.\\-_]+$`.
- Specification components pattern property regex needs to be valid against `u` flag.  This means changing `^x-[\\w\\d\\.\\-\\_]+$` to `^x-[\\w\\d.\\-_]+$`
- `pattern` property regex for parameter and correlationId `location` needs to be valid against `u` flag.  This means changing `^\\$message\\.(header|payload)\\#(\\/(([^\\/~])|(~[01]))*)*` to `^\\$message\\.(header|payload)#(\\/(([^\\/~])|(~[01]))*)*`
- Fixes `$ref` is incorrectly defined for a server.

**Related issue(s)**
Fixes https://github.com/asyncapi/spec-json-schemas/issues/164
Fixes #127 
Fixes https://github.com/asyncapi/spec-json-schemas/issues/120

**Paired PR's**
Spec adapting release process: https://github.com/asyncapi/spec/pull/659
Parser-js updating package and usage: https://github.com/asyncapi/parser-js/pull/423
Website exposing all these files remotely: https://github.com/asyncapi/website/pull/502


jonaslagoni:
There is one blocker, that I am not quite sure how we solve...

For this to work, the separate schemas need to be bundled together again to provide the full file once again, i.e. automatically bundle all the schemas together to form https://github.com/asyncapi/spec-json-schemas/blob/master/schemas/2.2.0.json. 

For now, I have not found any library yet, that accurately provides this behavior... I could implement it myself, as it is quite an easy hack, but I kinda don't want to...

Whenever we do find a solution, we need to create a workflow that automatically bundles the files when changes to the files have been made.

jonaslagoni:
As the implementation of a bundling behavior is quite simple, lets try to focus on https://github.com/hyperjump-io/json-schema-bundler

It however does have one blocking "bug" - https://github.com/hyperjump-io/json-schema-bundler/issues/2

magicmatatjahu:
@jonaslagoni did you try with https://github.com/APIDevTools/json-schema-ref-parser/blob/main/docs/ref-parser.md#bundleschema-options-callback ?

jonaslagoni:
> @jonaslagoni did you try with https://github.com/APIDevTools/json-schema-ref-parser/blob/main/docs/ref-parser.md#bundleschema-options-callback ?

hmm, I have not. Maybe this fits more our use-case 🤔

jonaslagoni:
Ignore the request, need to refactor some stuff to make parser compatible.

jonaslagoni:
So... The "problem" is that tooling has a very different way of handling the bundling, and I am not quite sure what to choose. 

I am not able to get any of the tools to replicate the full AsyncAPI file we currently have. This means that we need to force a major version change. i.e. [Our parser expects](https://github.com/asyncapi/parser-js/blob/511d39c3597c69f8b98284aad138396307c15968/lib/asyncapiSchemaFormatParser.js#L88) the schema object to be located under `#/definitions/schema`, with any of the underlying tools that are not possible.

1. [RefParser](https://github.com/APIDevTools/json-schema-ref-parser/blob/main/docs/ref-parser.md#bundleschema-options-callback) inlines everything. Making it one bundled mess, but would be possible to use.
2. [Hyperjump bundler](https://github.com/hyperjump-io/json-schema-bundler) only works with URI `$id` fields, making it incompatible with the current parser implementation.
This means we will have to do something like this:

```json
  "patternProperties": {
    "^x-[\\w\\d.\\-_]+$": {
      "$ref": "http://asyncapi.com/draft-07/schemas/specificationExtension"
    }
  },
```
```json
{
  "additionalItems": true,
  "$id": "http://asyncapi.com/draft-07/schemas/specificationExtension"
}
```
This will bundle it like:
```json
    "definitions": {
        "http://asyncapi.com/draft-07/schemas/specificationExtension": {
            "$id": "http://asyncapi.com/draft-07/schemas/specificationExtension",
            "description": "Any property starting with x- is valid.",
            "additionalProperties": true,
            "additionalItems": true
        },

    }
```

3. We could go with our "own" implementation and it would be quite "easy" to do 😄 But another tool, and as bundling behavior have [finally been formalized](https://json-schema.org/blog/posts/bundling-json-schema-compound-documents), we would be going against this logic.

So what do you think we should do? 🤔

definitions/2.2.0/APIKeyHTTPSecurityScheme.json
```diff
@@ -29,10 +29,10 @@
...
   "patternProperties": {
     "^x-[\\w\\d.\\-_]+$": {
-      "$ref": "specificationExtension.json"
+      "$ref": "http://asyncapi.com/draft-07/schemas/specificationExtension"
     }
   },
   "additionalProperties": false,
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "APIKeyHTTPSecurityScheme.json"
+  "$id": "http://asyncapi.com/draft-07/schemas/APIKeyHTTPSecurityScheme"
```
magicmatatjahu, line 37: It should be reversed to the `.../schemas/draft-07/...` because `draft-07` points to schema version, not to AsyncAPI version 😄

magicmatatjahu: On the other hand, it can stay because $id only work in JSON Schemes... Mondays 😅

jonaslagoni: Just an example URI to show the change 😄 Probably `http://asyncapi.com/schemas/APIKeyHTTPSecurityScheme` or `http://asyncapi.com/schemas/2.2.0/APIKeyHTTPSecurityScheme` would be more suitable.

magicmatatjahu: - The `http://asyncapi.com/schemas/APIKeyHTTPSecurityScheme` will point to a schema that will not be changed, so it is better to have some version.
- The `http://asyncapi.com/schemas/2.2.0/APIKeyHTTPSecurityScheme` points to the AsyncAPI version, to we will need to update (copy?) schemas on each release 😄

jonaslagoni: We just need to resolve https://github.com/asyncapi/spec-json-schemas/pull/128#issuecomment-981344308

smoya:
> So what do you think we should do? 🤔

Getting an idea of what all about takes more than expected, ergo my delay in giving a response.

You say that https://github.com/hyperjump-io/json-schema-bundle tool forces you to define `$id` as absolute URIs. It was confusing to me, so I'm gonna clarify some things.

You can still make use of URIs such as `/draft-07/schemas/info` for `$ref`, since the `$id` of the document will be used as base URI, meaning `http://json-schema.org/draft-07/schemas/info` will become the [Retrieval URI](https://json-schema.org/understanding-json-schema/structuring.html#retrieval-uri).
However, it is not possible to bundle external schemas into definitions with a name different than the value in their `$id` fields, or rather generating a UUID. You can either read this in their repo [README](https://github.com/hyperjump-io/json-schema-bundle#api) (`bundle` method options arg) or check code [here](https://github.com/hyperjump-io/json-schema-bundle/blob/main/lib/index.js#L50-L57).

Meaning, as you pointed out, the Parser-js won't be able to resolve `#/definitions/schema`. Instead, it would need to be changed to `#/definitions/http://asyncapi.com/draft-07/schemas/schemas`. In fact, I'm not sure if the tooling we use would register all definitions as external Ids, so maybe it will just be stored in the schemas registry as `http://asyncapi.com/draft-07/schemas/schemas`.

At this point, I'm wondering if using URIs when declaring definitions are like a good practice in JSON Schema. I couldn't find any statement or valid example in their docs. We could ask Jason (maintainer of that particular library) if it would make sense to allow other naming strategies.

On the other hand, if we don't care about introducing breaking changes, I don't see a problem having those full URIs on definitions.

smoya:
@jonaslagoni Something we could move forward asap is the fix for the regexes. This could go into a separate PR as it is already failing when using a validator such as https://json-schema.hyperjump.io/

jonaslagoni:
Blocked by https://github.com/hyperjump-io/json-schema-bundle/issues/5

derberg:
Screenshot 2021-12-09 at 18 10 38

[Karma](https://en.wikipedia.org/wiki/Karma)?

🤣

README.md
```diff
@@ -2,7 +2,7 @@
 
 # AsyncAPI
 
-This package provides all the versions of the AsyncAPI schema.
+This is a mono repository, which provides all the JSON Schemas documents for validating AsyncAPI documents.
```
alequetzalli, line 5:
```suggestion
This is a mono repository, which provides all the JSON Schema documents for validating AsyncAPI documents.
```

I believe this should be singular :)

README.md
```diff
@@ -37,3 +37,27 @@ const asyncapi = versions['1.1.0'];
...
 // Do something with the schema.
 ```
+
+## Repository structure
+This is the current project structure explained.
+- [./definitions](./definitions) - contain all the individual schemas that will automatically be bundled together to provide the schemas in [./schemas](./schemas).
+- [./tools/bundler](./tools/bundler) - is the tool that bundles all the individual schemas together.
+- [./schemas](./schemas) - contain all automatically bundled and complete schemas for each AsyncAPI version. These schemas should **NOT** be manually changed as they are automatically generated. Any changes should be done in [./definitions](./definitions).
+
+## Bundling of schemas
```
alequetzalli, line 47:
```suggestion
## Schema Bundling
```

README.md
```diff
@@ -37,3 +37,27 @@ const asyncapi = versions['1.1.0'];
...
+Changes should not be done manually to the schemas in [./schemas](./schemas), but instead be done in their individual definitions located in [./definitions](./definitions).
+
+These definitions are automatically bundled together on new releases through the `generate:assets` script, which runs the [bundler](./tools/bundler). 
+
+For example for [2.2.0](./definitions/2.2.0), the [bundler](./tools/bundler/index.js) starts with the [asyncapi.json](definitions/2.2.0/asyncapi.json) file and recursively go through all references (`$ref`) to create the [appropriate bundled version](./schemas/2.2.0.json).
```
alequetzalli, line 52:
```suggestion
For example, for [2.2.0](./definitions/2.2.0), the [bundler](./tools/bundler/index.js) starts with the [asyncapi.json](definitions/2.2.0/asyncapi.json) file and recursively goes through all references (`$ref`) to create the [appropriate bundled version](./schemas/2.2.0.json).
```

README.md
```diff
@@ -37,3 +37,27 @@ const asyncapi = versions['1.1.0'];
...
+### Creating a new version
+To create a new version, simply run the following command:
+```
+npm run startNewVersion --newVersion=x.x.x
+```
+Where `x.x.x` is the new version you want to create.
+
+The manual process of creating a new version is to:
+1. Duplicate the latest version (`y.y.y`) under definitions (so we have the correct base to make changes from). Rename the folder to the new version.
+2. Search and replace in the new duplicated folder for `definitions/y.y.y` and replace it with `definitions/x.x.x`, where `x.x.x` is the new version.
```
alequetzalli, line 63:
```suggestion
1. Duplicate the latest version (`y.y.y`) under definitions (so we have the correct base to make changes from). 
2. Rename the folder to the new version.
3. Search and replace in the new duplicated folder for `definitions/y.y.y` and replace it with `definitions/x.x.x`, where `x.x.x` is the new version.
```

migrations/Migrate to version 3.md
```diff
@@ -0,0 +1,35 @@
+# Migrating to version 3
+
+In version 3, only breaking change, is that the provided bundled schemas located under [schemas](../schemas) now use completely different `$id`s and definition namings.
```
alequetzalli, line 3:
```suggestion
In version 3, the only breaking change is that the provided bundled schemas located under [schemas](../schemas) now use completely different `$id`s and definition namings.
```

migrations/Migrate to version 3.md
```diff
@@ -0,0 +1,35 @@
...
+
+> If you do not manually access schemas under `definitions` you can upgrade the version without any problems.
```
alequetzalli, line 5:
```suggestion
> If you don't manually access schemas under `definitions`, you can upgrade the version without any problems.
```

migrations/Migrate to version 3.md
```diff
@@ -0,0 +1,35 @@
...
+
+Previously each definition would have it defined such as:
```
alequetzalli, line 7:
```suggestion
Previously each definition would have it defined as:
```

migrations/Migrate to version 3.md
```diff
@@ -0,0 +1,35 @@
...
+  "definitions": {
+    "Reference": { ... },
+    "ReferenceObject": { ... },
+    "info": { ... },
+    ...
+  }
+}
+```
+
+With the new naming approach those definitions would be called:
```
alequetzalli, line 21:
```suggestion
With the new naming approach, those definitions would be called:
```

alequetzalli (commented):
Left my review @jonaslagoni 😀👍🏽👍🏽

jonaslagoni:
> Left my review @jonaslagoni 😀👍🏽👍🏽

Thanks, @alequetzalli, great suggestions 👍

alequetzalli (approved):
LGTM 🚀

jonaslagoni:
Tagging code owners, @derberg @fmvilas. As the PR is ready to be merged, I am assuming that we are waiting for the new release of the spec to be completed to take a decision on this? 🤔

derberg:
I'm fully supporting the change but as you wrote, definitely not a good time to merge before 2.3 release

fmvilas:
Yeah, agree. Let's wait for v2.3.0 to ship and it should be fine.

magicmatatjahu:
@jonaslagoni Did you test schemas after bundling with ParserJS? Does everything work?

jonaslagoni:
> @jonaslagoni Did you test schemas after bundling with ParserJS? Does everything work?

Yep, see I prepared it here: https://github.com/asyncapi/parser-js/pull/423

jonaslagoni:
@smoya your change in https://github.com/asyncapi/spec-json-schemas/pull/142/commits/6b65cb890eb32d35d6a8d04a07e9a130326cbfdc I have not applied as I used `^x-[\\w\\d.\\-_]+$` and `^\\$message\\.(header|payload)#(\\/(([^\\/~])|(~[01]))*)*`. 

I am not sure which one to keep?

jonaslagoni:
I have adapted the PR so it now contains the 2.3.0 changes. I would suggest that once @derberg creates the `next-major` I change the target branch and it should be ready for reviews, anything else that needs to be done 🤔? 

cc codeowners @derberg @fmvilas

derberg (changes requested):
If https://github.com/jonaslagoni/spec-json-schemas/tree/feature/improve_maintainability/schemas are result of the bundling, product of the release, do we want to commit them into repo? shouldn't these be only in a package and on GitHub release as an additional assets?

Also can you explain why we do `http://asyncapi.com/definitions/2.3.0/specificationExtension.json` if in the end there is nothing really under given URL

Is this why you want to mark it as breaking change?

package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "test": "nyc mocha",
     "release": "semantic-release",
-    "generate:assets": "echo 'No additional assets need to be generated at the moment'",
+    "generate:assets": "cd tools/bundler && npm i && npm run bundle",
```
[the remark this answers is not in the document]
jonaslagoni: Good point, splitting it out 👍

README.md
```diff
@@ -37,3 +38,28 @@ const asyncapi = versions['1.1.0'];
...
+## Repository structure
+This is the current project structure explained.
+- [./definitions](./definitions) - contain all the individual schemas that will automatically be bundled together to provide the schemas in [./schemas](./schemas).
+- [./tools/bundler](./tools/bundler) - is the tool that bundles all the individual schemas together.
+- [./schemas](./schemas) - contain all automatically bundled and complete schemas for each AsyncAPI version. These schemas should **NOT** be manually changed as they are automatically generated. Any changes should be done in [./definitions](./definitions).
+
+## Schema Bundling
+Changes should not be done manually to the schemas in [./schemas](./schemas), but instead be done in their individual definitions located in [./definitions](./definitions).
+
+These definitions are automatically bundled together on new releases through the `generate:assets` script, which runs the [bundler](./tools/bundler).
```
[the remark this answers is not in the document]
jonaslagoni: Makes sense, changed 👍

jonaslagoni:
> If https://github.com/jonaslagoni/spec-json-schemas/tree/feature/improve_maintainability/schemas are result of the bundling, product of the release, do we want to commit them into repo? shouldn't these be only in a package and on GitHub release as an additional assets?

Technically no I guess 🤔 Let me try that approach, now that we have `npm run build` it also makes sense.

> Also can you explain why we do http://asyncapi.com/definitions/2.3.0/specificationExtension.json if in the end there is nothing really under the given URL.

@derberg I made a suggestion to expose them, see https://github.com/asyncapi/website/pull/502.

This also means, if the library you use supports fetching references, you don't need the bundled version, as they will be fetchable. So depends on your use-case what you need 😄 

> Is this why you want to mark it as breaking change?

Yes, changing the id's of the schemas is the breaking change 🙂

jonaslagoni:
> Technically no I guess 🤔 Let me try that approach, now that we have `npm run build` it also makes sense.

@derberg not quite sure how we should handle the Go library as it expects the schemas to be present, and I dont see a way it can use the bundle library 😅

smoya:
> 

You should update your documents with the changes I made on the regular expressions. As you can see here, your version is not compliant: https://regex101.com/r/tlOY2C/1

smoya:
> @smoya your change in [6b65cb8](https://github.com/asyncapi/spec-json-schemas/commit/6b65cb890eb32d35d6a8d04a07e9a130326cbfdc) I have not applied as I used `^x-[\\w\\d.\\-_]+$`.
> 
> I am not sure which one to keep?

You should update your documents with the changes I made to the regular expressions. As you can see here, your version is not compliant: https://regex101.com/r/tlOY2C/1

package.json
```diff
@@ -4,9 +4,13 @@
...
   "scripts": {
-    "test": "nyc mocha",
+    "test": "npm run build && nyc mocha",
+    "build": "npm run bundle",
     "release": "semantic-release",
-    "generate:assets": "echo 'No additional assets need to be generated at the moment'",
+    "generate:assets": "npm run build",
+    "prepublishOnly": "npm run build",
+    "bundle": "cd tools/bundler && npm i && npm run bundle",
+    "startNewVersion": "newVersion=$npm_config_new_version && latestVersion=$(ls -d ./definitions/* | sort -V -r | head -1 | xargs -n 1 basename) && [ -d ./definitions/${newVersion} ] && echo Directory ./definitions/$newVersion already exist and cannot be overwritten. Please create a different version. || (cp -R ./definitions/$latestVersion ./definitions/$newVersion && find ./definitions/$newVersion -name '*.json' -exec sed -i '' \"s+definitions/$latestVersion+definitions/$newVersion+g\" {} +)",
```
derberg, line 13: 😱 😆

derberg (commented):
Can you also configure release candidates setup for `next-major`?


as for your comment about Go package needs. Actually, we need to move it out, same with NPM package that stores schemas. Have a look at https://github.com/asyncapi/spec-json-schemas/issues/130

Or we end up with more and more complex monorepo that will require us to break current workflows approach we have. It is doable, we just need to forget about GH releases, and do tags like these folks for example https://github.com/actions/toolkit/tags but does it make sense. Easier for us is to do a split

jonaslagoni:
> Can you also configure release candidates setup for `next-major`?

Done 👍 

> as for your comment about Go package needs. Actually, we need to move it out, same with NPM package that stores schemas. Have a look at https://github.com/asyncapi/spec-json-schemas/issues/130
> Or we end up with more and more complex monorepo that will require us to break current workflows approach we have. It is doable, we just need to forget about GH releases, and do tags like these folks for example https://github.com/actions/toolkit/tags but does it make sense. Easier for us is to do a split

Do you want me to remove the Go package now and remove the bundled schema files? Or should that be in another PR?

We are generating the assets on releases, so all we need to do is removing Go package and bundled schema files.

package.json
jonaslagoni: Yeea, not sure if we should just leave it, or create something like a `script` folder? 🤔

derberg (approved):
@jonaslagoni this PR is complex enough, so do not remove Go package, especially that NPM package much also be moved out.

@dalelane @fmvilas anything from your side. I think we are ready to merge. It goes to release branch so any overlooked issues will not affect official packages anyway

package.json
derberg: I definitely recommend moving it out into separate script, it is scary, and in separate script you can add comments 😅

jonaslagoni: I'll make sure to split it out in another PR to simplify it 🙂

fmvilas (approved):
LGTM 👍

dalelane (approved):
looks good!

PR #133, opened 2021-12-02
feat: allow servers and channels to be defined as components

smoya:
**Description**

This PR is part of https://github.com/asyncapi/spec/issues/660.
It allows both `servers` and `channels` to be defined as (and referenced from) components, meaning the following will be now valid:

```yaml
servers:
  production:
    $ref: '#/components/servers/myserver'
channels:
  some/events:
    $ref: '#/components/channels/myChannel'
components:
  servers:
    myserver:
      url: "http://localhost.com:5000/ws"
      protocol: ws
  channels:
    myChannel:
      description: "mychannel"
```

I'm attaching the diff between this new schema `2.3.0` vs the previous `2.2.0` to show the changes:

```diff
33,36c33
<       "type": "object",
<       "additionalProperties": {
<         "$ref": "#/definitions/server"
<       }
---
>       "$ref": "#/definitions/servers"
59a57
>       
165,168c163,166
<       "required": [
<         "url",
<         "protocol"
<       ],
---
>       "oneOf" : [
>         { "required" : ["url", "protocol"] },
>         { "required" : ["$ref"] }
>       ],    
175a174,176
>         "$ref": {
>           "$ref": "#/definitions/ReferenceObject"
>         },
202a204,209
>     "servers": {
>       "type": "object",
>       "additionalProperties": {
>         "$ref": "#/definitions/server"
>       }
>     },
262a270,275
>         },
>         "servers": {
>           "$ref": "#/definitions/servers"
>         },
>         "channels": {
>           "$ref": "#/definitions/channels"
```

**Related issue(s)**
https://github.com/asyncapi/spec/issues/660


schemas/2.3.0.json
```diff
@@ -0,0 +1,1498 @@
...
+        }
+      }
+    },
+    "server": {
+      "type": "object",
+      "description": "An object representing a Server.",
+      "oneOf" : [
+        { "required" : ["url", "protocol"] },
+        { "required" : ["$ref"] }
+      ],
```
smoya, line 166: This allows us to set required properties based on the presence of `$ref`. Either we require `url` and `protocol`, or only `$ref` to a server schema.

schemas/2.3.0.json
```diff
@@ -0,0 +1,1498 @@
...
+          "format": "uri"
+        }
+      },
+      "patternProperties": {
+        "^x-[\\w\\d\\.\\-\\_]+$": {
+          "$ref": "#/definitions/specificationExtension"
+        }
+      }
+    },
+    "channelItem": {
```
magicmatatjahu, line 490: You missed to update Channel Object :)

schemas/2.3.0.json
magicmatatjahu: But with that schema we can also write sucha a:

```
$ref: ...
variables:
  - ...
  - ...
```

I know that we wanna have such a functionality to override properties, but atm we don't clarify it in the spec, so we should support only single `$ref` and full Server Object. This same should be done with Channel object. 

EDIT: I also see problem that we allow to define the `$ref` inside `component/server`. I think that we should allow it only in the `#/servers/{server}` (also in the `#/components/channels`), but I am opened to suggestion.

magicmatatjahu (changes requested):

schemas/2.3.0.json
smoya: Regarding overriding properties, it is possible to do with the rest so (afaik) so I don't see why we should not allow it for the new ones (for the sake of consistency).

Also this PR is not aiming to introduce a breaking change so removing features, even though they are not named or specified in the spec, is not desired IMHO.

schemas/2.3.0.json
smoya: What should I add in there? It is already possible to use `$ref`.

magicmatatjahu: Ahh ok, I didn't see that.

schemas/2.3.0.json
magicmatatjahu: Ok I saw rest of the occurrences of `ReferenceObject` and you're right 😂  Feature that we have but nobody know about that, I like that moment.

But also with current solution you cannot write:

```yaml
$ref: ... 
url: ...
protocol: ...
```

and I'm not sure if someone should have possibility to override url/protocol, or shouldn't 🤔

smoya: > But also with current solution you cannot write:

You mean with `2.2.0` right? Indeed you can't. That's exactly what this PR introduces.
And I think we should give total freedom on what the users wants to do, if they want to override the protocol, URL, etc its on their own (yes, we don't officially support that in our spec)

magicmatatjahu: No, I mean your PR. `oneOf` works in this way that only one item in the array must be valid, so if you pass:

```yaml
$ref: ... 
url: ...
protocol: ...
```

you exactly follow by two schemas in `oneOf`, so your data/input is invalid.

You should change `oneOf` to the `anyOf` and it exactly means `at least one schema is valid`, not `only one schema is valid` :)

smoya: That's a really good catch 👍  Fixed!

magicmatatjahu:
@smoya You should wait for merge of it https://github.com/asyncapi/spec-json-schemas/pull/139 and then switch PR's destination to the release-branch :)

smoya:
> then switch PR's destination to the release-branch

The destination branch is already there and this PR is targeting to be merged into it. If there are no other changes to introduce in `2.3.0` that are ready to merge on this repo, I think we can proceed with merging this one.

magicmatatjahu:
hah, I got confused, sorry 😅

magicmatatjahu (approved):
LGTM

derberg:
looks good but we need to wait for https://github.com/asyncapi/spec-json-schemas/pull/139 and then you will need to solve some conflicts

smoya:
Would you mind approving this PR anyway? i won't mark it as ready to be merged until https://github.com/asyncapi/spec-json-schemas/pull/139 is merged.

derberg:
my approval won't make much sense as you will anyway have conflicts, which means additional commits that will cause a dismiss of my approval 😄

smoya:
> my approval won't make much sense as you will anyway have conflicts, which means additional commits that will cause a dismiss of my approval 😄

But it means a lot to my ❤️ , @derberg. Kidding, yeah you are right. Let's wait for those conflicts to come and get resolved!

smoya:
This branch is now up-to-date with `2022-01-release` and all conflicts have been solved. cc @derberg

derberg (approved):
/dnm

fmvilas (approved):

derberg:
https://github.com/asyncapi/spec/pull/665 merged, so merging this one too

/rtm

PR #136, opened 2021-12-08
feat: add the 2.3.0.json version, and add the Solace binding.

damaru-inc:
**Description**
This adds a new version 2.3.0.json and adds the solace binding to the list of bindings.

**Related issue(s)**
The related PR on the spec itself is here: https://github.com/asyncapi/spec/pull/666

### DIFF
--- a/.github/workflows/automerge-for-humans-add-ready-to-merge-or-do-not-merge-label.yml
+++ b/.github/workflows/automerge-for-humans-add-ready-to-merge-or-do-not-merge-label.yml
@@ -8,7 +8,7 @@ on: issue_comment
 
 jobs:
   parse-comment-and-add-ready: # for handling cases when you want to mark as ready to merge
-    if: github.event.issue.pull_request && github.event.issue.state != 'closed'
+    if: github.event.issue.pull_request && github.event.issue.state != 'closed' && github.actor != 'asyncapi-bot'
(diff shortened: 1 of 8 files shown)


derberg (changes requested):
@MichaelDavisSolace sorry but unfortunately you will have to recreate this PR, like create new, unless you are a git expert and can handle it here.

Basically problem is that you cat your branch from `master` and not release branch and you are now pushing lots of stuff not related to your PR. We should have only your `solace` binding related things.

Also to better handle next contribs, we need to let https://github.com/asyncapi/spec-json-schemas/pull/139 in tomorrow first and this will for sure cause some conflicts in your PR, sorry about that :(

damaru-inc:
> @MichaelDavisSolace sorry but unfortunately you will have to recreate this PR, like create new, unless you are a git expert and can handle it here.
> 
> Basically problem is that you cat your branch from `master` and not release branch and you are now pushing lots of stuff not related to your PR. We should have only your `solace` binding related things.
> 
> Also to better handle next contribs, we need to let #139 in tomorrow first and this will for sure cause some conflicts in your PR, sorry about that :(

No problem!

damaru-inc:
Replaced by [143](https://github.com/asyncapi/spec-json-schemas/pull/143), which branches off the feature branch.

Issue #130, opened 2021-11-24
remove js and go package and probably move json schemas should stay alone in the repo


derberg:
https://github.com/asyncapi/spec-json-schemas/pull/124 introduced symetric releases for the package. So basically whenever there is a change in Go package, we will also release JS package in all the different projects that depend on it. Cause a lot of noise with "empty" functionality.

1. Go package should have its own release cycle. Symetric releases are cool but for cases like in react-component where we release web-component and other things at the same time
1. I know some time ago we took JSON Schema files out from `spec` repo and moved it here to have one project with JSON Schemas and projects for different languages. But I admit, that in this discussion I never took into consideration release process and how strange things will work here. IMHO the best would be to leave JSON Schemas in this repo alone, then JS code should go to `js-json-schemas` and Go should go to something like `go-json-schemas`. Once JSON Schema is released with some update, there should be a workflow that pushes updated json schema to JS, Go and other projects and release is triggered there automatically

Related conversation: https://github.com/asyncapi/spec-json-schemas/issues/118

derberg:
@fmvilas @smoya have a look

smoya:
I wasn't very conscious about the fact we have so many dependents on this right package: https://github.com/asyncapi/spec-json-schemas/network/dependents.
Releasing a new version of it, without real changes that affect the package you depend on, is definitely not something ideal; such as getting an update of the node package triggered by a Go code change.

On the other hand, the idea of having all language implementations in the same repository introduces few interesting benefits: 

- Test suites for each language implementation run on every PR that introduces changes on the schemas, meaning whatever you merge, it is supported by all different implementations. I have to say that with current implementations, I would not expect something to get broken whenever we introduce changes on the schemas (we mostly create new files, right?) but not sure how each package will evolve in time.
- Packages for each language implementation are automatically released whenever we introduced a change on the schemas VS having to create a PR on each implementation repository, review it, accept it and merge it (feature disparity will happen across languages). We could automate this? I think so, but it is an extra effort and could eventually fail (tests could fail, related to what I mentioned earlier).

It is fair if we want to think deeply about it and decide what do we want to leave behind, as as far as I can see, there is no silver bullet.

fmvilas:
Can't we have a monorepo in this case? I mean, what would we need to make releases work depending on the directory that has been modified? Just curious.

smoya:
[...]

PR #138, opened 2021-12-09
chore: add CODEOWNERS

derberg:
@fmvilas I wasn't sure here so I took an approach that JSON Schema is very closely aligned with `spec` then it should have exactly the same owners as `spec` repo.

We can rethink once we make a final decision about https://github.com/asyncapi/spec-json-schemas/issues/130

Thoughts?

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,9 @@
+# This file provides an overview of code owners in this repository.
+
+# Each line is a file pattern followed by one or more owners.
+# The last matching pattern has the most precedence.
+# For more details, read the following article on GitHub: https://help.github.com/articles/about-codeowners/.
+
+# The default owners are automatically added as reviewers when you open a pull request unless different owners are specified in the file.
+
+* @fmvilas @derberg @github-actions[bot]


CODEOWNERS
```diff
@@ -0,0 +1,9 @@
+# This file provides an overview of code owners in this repository.
+
+# Each line is a file pattern followed by one or more owners.
+# The last matching pattern has the most precedence.
+# For more details, read the following article on GitHub: https://help.github.com/articles/about-codeowners/.
+
+# The default owners are automatically added as reviewers when you open a pull request unless different owners are specified in the file.
+
+* @fmvilas @derberg @github-actions[bot]
```
fmvilas, line 9:
```suggestion
* @fmvilas @derberg @jonaslagoni @github-actions[bot]
```

fmvilas (changes requested):
Sounds good. My only suggestion is to add @jonaslagoni because of this: https://github.com/asyncapi/spec-json-schemas/pull/128. I know it's not yet merged but that solves a lot of bugs and improves the repo by a lot.

CODEOWNERS
derberg: but then we are not aligned with `spec` repo 😉

derberg: so maybe let us get #130 and #128 done first and then we can add more folks, huh?

fmvilas: Yeah, I'm happy to wait but I'd not worry about aligning with spec. JSON Schemas are not the spec but a partial representation of it. If we align here we'd have to align on parsers too, which I don't think makes sense. What do you think?

derberg: so I know that we treat JSON Schema as a tool, but I was thinking about the alignment because it almost in 90% reflects the spec, not like in case of parser, and parser anyway depends on JSON Schema again in 90% for the validation.

if `spec` owners are not only gatekeepers for JSON Schema, in theory, it may happen that some validation rule can be changed here but not in spec. You know what I mean?

🤔

fmvilas: > if spec owners are not only gatekeepers for JSON Schema, in theory, it may happen that some validation rule can be changed here but not in spec. You know what I mean?

Yeah, I get it, I just don't think it's very different than a parser implementing a custom validation that is not defined in the spec (like in https://github.com/asyncapi/spec/issues/650). Anyway, I'm not going to block this PR just because of that.

I'm approving and we can consider adding other people later 👍

fmvilas (approved):

PR #145, opened 2022-01-04
fix: make pattern and patternProperties unicode ECMA-262 compliant

smoya:
**Description**

This PR makes the regular expressions found in `pattern` and `patternProperties` fields unicode compliant (according to ECMA-262). It only changes the `2.3.0.json` schema but if you consider this should be applied to the rest, I can expand it.

It means now the regular expressions are compatible with `u` flag, flag that some JSON Schema parsers use when parsing a JSON Schema document. For example the one from Hyperjump](https://github.com/hyperjump-io/json-schema-validator). See implementation [here](https://github.com/hyperjump-io/json-schema-validator/blob/521818176d71d82344450a40a405fff2029e6bcc/lib/keywords/patternProperties.js) and [here}(https://github.com/hyperjump-io/json-schema-validator/blob/521818176d71d82344450a40a405fff2029e6bcc/lib/keywords/pattern.js). This is being used by the online tool https://json-schema.hyperjump.io.

This is not introducing a BC since the regular expressions didn't change their meaning. You can see each version with each regex explanation on the following links:

- patternProperties: 
  - [old](https://regex101.com/r/xhn0Bf/2)
  - [new](https://regex101.com/r/nEJXSz/1)
- pattern
  - [old](https://regex101.com/r/Ny2oXU/1)
  - [new](https://regex101.com/r/q7cgyA/1)

**Related issue(s)**
Based on https://github.com/asyncapi/spec-json-schemas/pull/128 but fixing the proposed for the `patternProperties` with a proper unicode character.

cc @jonaslagoni @fmvilas

### DIFF
--- a/schemas/2.3.0.json
+++ b/schemas/2.3.0.json
@@ -9,7 +9,7 @@
   ],
   "additionalProperties": false,
   "patternProperties": {
-    "^x-[\\w\\d\\.\\-\\_]+$": {
+    "^x-[\\w\\d\\.\\x2d_]+$": {
       "$ref": "#/definitions/specificationExtension"
     }
   },
@@ -80,7 +80,7 @@
       ],
       "additionalProperties": false,
       "patternProperties": {
-        "^x-[\\w\\d\\.\\-\\_]+$": {
+        "^x-[\\w\\d\\.\\x2d_]+$": {
           "$ref": "#/definitions/specificationExtension"
         }
       },
@@ -131,7 +131,7 @@
         }
       },
       "patternProperties": {
-        "^x-[\\w\\d\\.\\-\\_]+$": {
+        "^x-[\\w\\d\\.\\x2d_]+$": {
           "$ref": "#/definitions/specificationExtension"
         }
       }
@@ -154,7 +154,7 @@
         }
       },
       "patternProperties": {
-        "^x-[\\w\\d\\.\\-\\_]+$": {
+        "^x-[\\w\\d\\.\\x2d_]+$": {
           "$ref": "#/definitions/specificationExtension"
         }
       }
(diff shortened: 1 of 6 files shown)


schemas/2.3.0.json
```diff
@@ -9,7 +9,7 @@
   ],
   "additionalProperties": false,
   "patternProperties": {
-    "^x-[\\w\\d\\.\\-\\_]+$": {
+    "^x-[\\w\\d\\.\\x2d_]+$": {
```
jonaslagoni, line 12: Good catch! 

Hmmm 🤔 

I guess the problem is because it is inside a matching list `[...]` 🤔 

Otherwise I guess we needed to do: `     "^x\\x2d[\\w\\d\\.\\x2d_]+$": {`

jonaslagoni (approved):
LGTM, as it solves the problem 👍 

Weird that the implementation of hyperjump actually supports `^x-[\\w\\d.\\-_]+$` but proper regex implementation does not 😅 

Only had one comment, but guess regex behaves differently in different scenarios 🤷

schemas/2.3.0.json
smoya: Yes, you guessed correctly. It is because that `-` is inside a list.

smoya:
> Weird that the implementation of hyperjump actually supports `^x-[\\w\\d.\\-_]+$` but proper regex implementation does not 😅

It is because it uses an old version of the validator:

```
"@hyperjump/oas-schema-validator": "^0.5.0",
```

The `u` flag was not added by then in the code. See https://github.com/hyperjump-io/json-schema-validator/blob/v0.5.0/lib/keywords/patternProperties.js

fmvilas:
> It only changes the 2.3.0.json schema but if you consider this should be applied to the rest, I can expand it.

If it's not breaking anything and it's actually a bug fix, I think this should be applied to all the files.

smoya:
> If it's not breaking anything and it's actually a bug fix, I think this should be applied to all the files.

I did push a commit updating the `2.x` schemas. I skipped `1.x` because patterns are way different.

fmvilas (approved):

Issue #82, opened 2021-07-29
Validate `object` type for the headers in the message trait 


magicmatatjahu:
As we read in the specification about `headers` field for Message Object: 

![image](https://user-images.githubusercontent.com/20404945/127496940-ba5757ce-068b-4bb8-a710-5bc0a38558f1.png)

we validate it by the lines:

https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.1.0.json#L644-L657

However we should also validate the `headers` in the Message Trait Object: 

https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.1.0.json#L885-L892

JSON Schemas for `2.0.0` and `2.1.0` should be updated.

PR #155, opened 2022-01-18
fix: force object type for message trait's headers

magicmatatjahu:
**Description**

Force object type for message trait's headers.

**Related issue(s)**
Part of #82

### DIFF
--- a/schemas/2.3.0.json
+++ b/schemas/2.3.0.json
@@ -904,12 +904,16 @@
           "type": "string"
         },
         "headers": {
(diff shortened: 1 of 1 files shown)


dalelane:
@derberg @fmvilas Are you happy for this to go into 2.3? If so, are you able to review and merge please?

fmvilas (approved):
Yeah looks good to me as the spec already says it MUST be of type "object" 👍

derberg (approved):

PR #171, opened 2022-02-09
feat: let channels be identified by ID. Address is now moved into a new field

smoya:
This PR allows channels to be identified by an ID rather than by their address. Instead, the address has now been moved into a new field called `address`.

This is a breaking change so it should be considered as a candidate for `v3.0.0`.

**Related issue(s):**

See https://github.com/asyncapi/spec/issues/663 for knowing about the reason behind this RFC.

CC @char0n @magicmatatjahu @derberg

### DIFF
--- a/definitions/3.0.0/channelItem.json
+++ b/definitions/3.0.0/channelItem.json
@@ -1,5 +1,8 @@
 {
   "type": "object",
+  "required": [
+    "address"
+  ],
   "additionalProperties": false,
   "patternProperties": {
     "^x-[\\w\\d\\.\\x2d_]+$": {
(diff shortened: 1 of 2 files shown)


smoya:
@fmvilas @derberg would you mind removing the `stale` label?

jonaslagoni:
@smoya if you switch branch you can convert this 🙂

smoya:
> @smoya if you switch branch you can convert this to the new definitions 🙂

Yeah, thanks. The reality is that maintaining a change in a new JSON Schema file (3.0) for a long time (major version) is hard since other versions will happen in the meantime. Keeping the new schema up to date is a manual and hard task.

smoya:
Closing in favour of https://github.com/asyncapi/spec-json-schemas/pull/264

PR #181, opened 2022-03-14
refactor: add script to add new version

jonaslagoni:
**Description**
As discussed in #128 this PR adds a new script to make it more maintainable instead of one long shell script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "generate:assets": "npm run build",
     "prepublishOnly": "npm run build",
     "bundle": "cd tools/bundler && npm i && npm run bundle",
-    "startNewVersion": "newVersion=$npm_config_new_version && latestVersion=$(ls -d ./definitions/* | sort -V -r | head -1 | xargs -n 1 basename) && [ -d ./definitions/${newVersion} ] && echo Directory ./definitions/$newVersion already exist and cannot be overwritten. Please create a different version. || (cp -R ./definitions/$latestVersion ./definitions/$newVersion && find ./definitions/$newVersion -name '*.json' -exec sed -i '' \"s+definitions/$latestVersion+definitions/$newVersion+g\" {} +)",
+    "startNewVersion": "newVersion=$npm_config_new_version node scripts/add-new-version.js",
     "lint": "echo 'No linter integrated yet'",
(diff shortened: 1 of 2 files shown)


jonaslagoni:
I have no idea what the sonar rules are called, so I don't know how to disable them.

I did add a simple input check to see if it matches semver version, so there should be no unsafe OS commands possible.

For the codesmell, cant really do anything about it, as this is the regex suggested on the semver site.

derberg:
@jonaslagoni sonar is complaining about script that runs only in CI, so just ignore it entirely for sonar

I have other problem with this PR, you make many changes that come from the fact that you create a PR not from `next-major` but from `master`, and therefore there are these workflow changes that should not be there. You know what I mean?

jonaslagoni:
> I have other problem with this PR, you make many changes that come from the fact that you create a PR not from `next-major` but from `master`, and therefore there are these workflow changes that should not be there. You know what I mean?

Pr is up https://github.com/asyncapi/spec-json-schemas/pull/187

jonaslagoni:
> @jonaslagoni sonar is complaining about script that runs only in CI, so just ignore it entirely for sonar

Is this something we have to do from the platform itself or? 🤔

derberg:
@jonaslagoni no, we always tend to do it with config files in repo, like before...or I didn't get something

btw, you need to update PR with latest `next-major`

jonaslagoni:
> btw, you need to update PR with latest `next-major`

I have.

> @jonaslagoni no, we always tend to do it with config files in repo, like before...or I didn't get something

Problem is, as I think it was last time. I dont have access to the configuration "keys" that is needed to disable the check 🤔

derberg:
sorry, I thought you are aware of something like https://github.com/asyncapi/glee/blob/master/.sonarcloud.properties so you can exclude entire script from all the rules

oko, so now we have a script, but how is it triggered?

jonaslagoni:
Gonna redo this once https://github.com/asyncapi/spec-json-schemas/pull/184 is merged.

jonaslagoni:
Damn, when targeted branches are removed, remaining PRs are closed as well with no way to retarget them...

PR #185, opened 2022-03-21
feat: release for version 2.4.0 of the spec

smoya:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@asyncapi/specs",
-  "version": "2.13.2",
+  "version": "2.14.0-2022-04-release.1",
   "description": "AsyncAPI schema versions",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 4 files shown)


derberg (approved):

smoya:
@dalelane @fmvilas @derberg need a +1 on this + remove `do-not-merge` label so we can merge and release it 🙏

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@asyncapi/specs",
-  "version": "2.13.2",
+  "version": "2.14.0-2022-04-release.3",
```
magicmatatjahu, line 3: This should be probably changed to the `2.13.2`

derberg: no, RC should stay, `semantic-release` will figure it out, notice 2.13 and bump to 2.14 🍻

magicmatatjahu: Great ❤️

smoya:
Only @dalelane approval is left and Ok to go!

PR #189, opened 2022-03-25
feat: security support added to operation and operationTrait 

LokeshRishi:
2 passing (789ms)

----------|---------|----------|---------|---------|-------------------
File      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
----------|---------|----------|---------|---------|-------------------
All files |     100 |      100 |     100 |     100 |                   
 index.js |     100 |      100 |     100 |     100 |                   
----------|---------|----------|---------|---------|-------------------

**Description**
Updates to include 'security' at operation and operationTrait level.

**Related issue(s)**
https://github.com/asyncapi/spec/pull/584

### DIFF
--- a/schemas/2.4.0.json
+++ b/schemas/2.4.0.json
@@ -616,6 +616,12 @@
         "operationId": {
           "type": "string"
         },
+        "security": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/SecurityRequirement"
+          }
+        },
         "bindings": {
           "$ref": "#/definitions/bindingsObject"
         },
@@ -869,6 +875,12 @@
         },
         "description": {
           "type": "string"
+        },
+        "security": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/SecurityRequirement"
+          }
         },
         "tags": {
           "type": "array",


smoya:
Thanks for the PR @LokeshRishi! would you mind changing the title of this PR adding a prefix to be compliant with [Conventional Commits](https://www.conventionalcommits.org)? `feat:` prefix seems the right one.

schemas/2.4.0.json
```diff
@@ -616,6 +616,12 @@
         "operationId": {
           "type": "string"
         },
+        "security": {
```
smoya, line 619: Would it make sense to extract this into a new definition called `SecurityRequirements`? Then you will just need to use it by `"$ref": "#/definitions/SecurityRequirements"`

sekharbans-ebay: > Would it make sense to extract this into a new definition called `SecurityRequirements` or similar? Then you will just need to use it by `"$ref": "#/definitions/SecurityRequirements"`

Hi @smoya  - Its an array of SecurityRequirements. Not sure if just indicating an array of a complex schema warrants a new definition. It would apply to all similar constructs (tags for example). Not sure it serves much purpose.

smoya: Fair enough 👍 I agree we would need to be consistent and lot of changes will need to be applied.

smoya (approved):
LGTM, however we need a +1 of a Code Owner.

dalelane (approved):
LGTM

derberg (approved):
@fmvilas only your approval is missing

I put https://github.com/asyncapi/spec/pull/584 on Stage 2, 
once we merge this and parser PR, then we can set Stage 3 and merge

PR #208, opened 2022-04-21
feat!: release for version 3.0.0 of the spec

jonaslagoni:

### DIFF
--- a/.github/workflows/automerge-for-humans-add-ready-to-merge-or-do-not-merge-label.yml
+++ b/.github/workflows/automerge-for-humans-add-ready-to-merge-or-do-not-merge-label.yml
@@ -50,7 +50,7 @@ jobs:
               repo: pull.head.repo.name,
               basehead: `${pull.base.label}...${pull.head.label}`,
             });
-            if (comparison.behind_by !== 0) {
+            if (comparison.behind_by !== 0 && pull.mergeable_state === 'behind') {
               console.log(`This branch is behind the target by ${comparison.behind_by} commits`)
               console.log('adding out-of-date comment...');
               github.rest.issues.createComment({
--- a/.github/workflows/automerge-orphans.yml
+++ b/.github/workflows/automerge-orphans.yml
@@ -57,7 +57,7 @@ jobs:
         name: Send info about orphan to slack
         uses: rtCamp/action-slack-notify@v2
         env:
-          SLACK_WEBHOOK: ${{secrets.SLACK_GITHUB_NEWISSUEPR}}
+          SLACK_WEBHOOK: ${{secrets.SLACK_CI_FAIL_NOTIFY}}
           SLACK_TITLE: 🚨 Not merged PR that should be automerged 🚨
           SLACK_MESSAGE: ${{steps.issuemarkdown.outputs.text}}
           MSG_MINIMAL: true
--- a/.github/workflows/link-check-cron.yml
+++ b/.github/workflows/link-check-cron.yml
@@ -29,8 +29,8 @@ jobs:
     - name: Report workflow run status to Slack
       uses: 8398a7/action-slack@v3
       with:
-          status: ${{ job.status }}
-          fields: repo,message,action,eventName,ref,workflow
+        status: ${{ job.status }}
+        fields: repo,action,workflow
       env:
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_DOCS_CHANNEL }}
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_CI_FAIL_NOTIFY }}
       if: failure() # Only, on failure, send a message on the Slack Docs Channel (if there are broken links)
--- a/.github/workflows/link-check-pr.yml
+++ b/.github/workflows/link-check-pr.yml
@@ -6,12 +6,15 @@ name: Check Markdown links
 on: 
   pull_request_target:
     types: [synchronize, ready_for_review, opened, reopened]
+    paths:
+      - '**.md'
 
 jobs:
   External-link-validation-on-PR:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - name: Checkout repo
+      uses: actions/checkout@v3
     - name: Check links
       uses: gaurav-nelson/github-action-markdown-link-check@v1
       with:
--- a/package.json
+++ b/package.json
@@ -48,6 +48,10 @@
       {
         "name": "2022-04-release",
         "prerelease": true
+      },
+      {
+        "name": "next-major-spec",
+        "prerelease": true
       }
     ],
     "plugins": [


smoya:
After fixing the Sonarcloud code smells, we can proceed with the merge of this PR which will trigger a new major release.

cc @jonaslagoni @derberg @fmvilas @dalelane @char0n

derberg:
Integration of bindings schemas will break a lot of AsyncAPI documents for users, this should be considered a breaking change

smoya:
I changed the title prefix to become `feat!` (with exclamation mark) so this is considered as a BC because of the bindings inclusion @derberg mentioned above.

We are always in time to revert the title if needed.

cc @jonaslagoni

jonaslagoni:
> Integration of bindings schemas will break a lot of AsyncAPI documents for users, this should be considered a breaking change

@derberg but this is only related to v3, that is not yet released?

derberg:
@jonaslagoni ah, sorry, I did not check and forgot that you integrated bindings schemas only with v3 and not previous schemas. Then yeah, makes sense why it is not a breaking change, and we can actually even release it without waiting for v3 spec 🤔

jonaslagoni:
I think we should go ahead and release this ASAP and treat the AsyncAPI v3 schema as a feature flag (using it can mean breaking changes without needing major version), the same way we handle it for the converter at the moment.

What do you folks think?

derberg:
what is the difference between releasing it as minor with feature flag vs just using release candidate? so better just integrate with RC imho

jonaslagoni:
More maintenance work with keeping branches up to date, and it's hard to release everything at once when the final release of v3 need to happen i.e. long sync meetings (as in the past) to release the `next-major-spec` branches. Seeing the v3 input as a feature flag solves that, cause you don't have to change anything when we release v3 in the spec.

derberg:
oh yeah, this is a definitely a good reason to release early, I agree 100%

what were your thoughts about feature flag? because you know that not all people use `@asyncapi/spec`, there are people that for other languages just take schemas directly.

what if we just in API do something like `'3.0.0-alpha': require('./schemas/3.0.0-alpha.json'),` and in docs also write that it is a subject to frequent change?
this way we have it pretty clear for person that tries to use it on API level, or directly.

jonaslagoni:
The most optimal from our point of view would IMO be a solution that require no release at all. I.e. indirectly when using 3.0.0 documents means the schemas might contain breaking changes from spec-json-schema version to version.

If any other solution, then you are back to the same problem meaning everything has to be released at the same time. Although you could of course minimize what needs to change, as you suggested here.

The problem is nowhere are we using `3.0.0-alpha` or letting the user know by reading https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.12 that this is a thing 🤷 So would create more confusion IMO.

derberg:
> no release at all. I.e. indirectly when using 3.0.0 documents means the schemas might contain breaking changes from spec-json-schema version to version.

what do you mean no release? sorry, I didn't get that 🤔 
not sure what are you proposing

jonaslagoni:
> what do you mean no release? sorry, I didn't get that 🤔
> not sure what are you proposing

That we release as and if you use (access) `'3.0.0': require('./schemas/3.0.0.json'),` it implicitly means these schemas CAN contain breaking changes until the release of v3. The same principle applied to the converter, that if you use `convert('3.0.0', ...)` that can contain breaking changes between versions until the release of v3.

That means we don't have to change anything in the spec-json-schemas repo when we release v3.

derberg:
> it implicitly means these schemas CAN contain breaking changes until the release of v3

it is a pretty huge assumption 😃 
JSON Schema is just a tool, not part of the spec. The fact that spec is not released do not implicitly mean there are breaking changes coming to JSON Schema. This repo is already released with majors when breaking change is coming in. As a user I would expect new major when there is a breaking change in JSON Schema for v3

So I'm all in to release as it is, as minor for now, but if in 2 weeks we need to introduce a breaking change, we should release major.

> The problem is nowhere are we using 3.0.0-alpha

then lets call it `3.0.0-Pre-release` as in the spec, and this way we are clear and consistent

jonaslagoni:
>As a user I would expect new major when there is a breaking change in JSON Schema for v3

When I say breaking change, what I mean is that the underlying schemas can contain breaking changes based on the current development of AsyncAPI v3, ONLY v3, not v2 or any set-in-stone versions - And I am not referring to the API of the library unless it's directly related to the usage of `3.0.0`.

This approach means we can entirely avoid long-lived branches across all tools if we adopt this approach.

The overall philosophy behind this approach is, if you use an AsyncAPI version in tools that have not been released yet, it can contain breaking changes until it is. Indirectly using any v3-related features as an unstable feature-flag until the spec is released.

> then lets call it 3.0.0-Pre-release as in the spec, and this way we are clear and consistent

Then you have to adapt the `asyncapi` property as well, as well in the spec, otherwise, you are required to do custom conditions in your tools when looking up the schema. 

Normally tools use the `.asyncapi` directly to fetch the appropriate schema.

It just complicates things, then we might as well keep the long-lived branch and its released version and do one major merge in the end.

derberg:
so when you write

> and treat the AsyncAPI v3 schema as a feature flag

you do not mean a real feature flag that someone has to pass?

my understanding is that your solution aims not only for v3 but also future releases to get rid of long living branches - **and I like it a lot**. 

I just don't think it is a good DX that we extend below with just `3.0.0`:
```
module.exports = {
    'schemas': {
        '2.0.0': require('./schemas/2.0.0.json'),
        '2.1.0': require('./schemas/2.1.0.json'),
        '2.2.0': require('./schemas/2.2.0.json'),
        '2.3.0': require('./schemas/2.3.0.json'),
        '2.4.0': require('./schemas/2.4.0.json'),
        '2.5.0': require('./schemas/2.5.0.json'),
        '2.6.0': require('./schemas/2.6.0.json'),
    },
    'schemasWithoutId': {
        '2.0.0': require('./schemas/2.0.0-without-$id.json'),
        '2.1.0': require('./schemas/2.1.0-without-$id.json'),
        '2.2.0': require('./schemas/2.2.0-without-$id.json'),
        '2.3.0': require('./schemas/2.3.0-without-$id.json'),
        '2.4.0': require('./schemas/2.4.0-without-$id.json'),
        '2.5.0': require('./schemas/2.5.0-without-$id.json'),
        '2.6.0': require('./schemas/2.6.0-without-$id.json'),
    }
};
```

and it is by definition unstable, and you learn about it only from readme. And what if you are long time user?

what about:
```
module.exports = {
    'schemas': {
      ...
    },
    'schemasWithoutId': {
      ...
    },
    'preReleases': {
      ...
     },
    'preReleasesWithoutId': {
      ...
     }
};
```

---
Alternative:
- we make it clear, in main readme how the releases work here, that pre releases are available, etc
- we not release as new major, and in the GitHub Release we clearly specify what is the breaking change and how future releases will behave

This way we make sure current users of our schemas are secure, and when they move to new major, they actually read the note

jonaslagoni:
>Alternative:
> - we make it clear, in main readme how the releases work here, that pre releases are available, etc
> - we not release as new major, and in the GitHub Release we clearly specify what is the breaking change and how future releases will behave
> This way we make sure current users of our schemas are secure, and when they move to new major, they actually read the note

I like the alternative way best because it means that once we release v3 we only need to merge spec and website without having to bump or change anything else 😄

But yea, we could do it as an experiment, see whether anything bad gets back to us, and then we can revert and adapt the implementation to your above suggestion. How does that sound?  @derberg

derberg:
> I like the alternative way best because it means that once we release v3 we only need to merge spec and website without having to bump or change anything else 😄

yeah, since I wrote it (kinda verbalised) I like it too :) 

> But yea, we could do it as an experiment, see whether anything bad gets back to us, and then we can revert and adapt the implementation to your above suggestion

wait, you mean release intentionally a potential breaking change and wait for explosion?

jonaslagoni:
> wait, you mean release intentionally a potential breaking change and wait for explosion?

@derberg only for v3 schemas, for example if we had merged this PR before #417 was merged, that PR would technically have been a breaking change (or fix I guess, dependant on your perspective), but because it's v3 targeted, we would have merged it as a feature or fix, because its for a pre-release version.

derberg:
> @derberg only for v3 schemas, for example if we had merged this PR before https://github.com/asyncapi/spec-json-schemas/pull/417 was merged, that PR would technically have been a breaking change (or fix I guess, dependant on your perspective), but because it's v3 targeted, we would have merged it as a feature or fix, because its for a pre-release version.

yes, first merge is not a problem, but all the potential breaking-changes followup merges without major bump are

> we could do it as an experiment, see whether anything bad gets back to us

what you are suggesting is, we release as minor, let people use it, and then potentially merge a breaking change and see what might go wrong? if people complain.

I don't think it is ok. We need either major with proper note or minor with https://github.com/asyncapi/spec-json-schemas/blob/master/index.js modification to add another properties, next to `schemas` and `schemasWithoutId` that hold RC versions. Then it is up to the user to decide if they want to adjust their code to support RC versions or not.

as you noted already I think, for example our parser-js, just reads `schemas` and out of the box supports all versions that are there. So if we do it, others do it probably too. So by automatically bumping new minor, they might turn on 3.0 support not knowing, and their users might generate complain traffic that something do not work. We should definitely not experiment this way

derberg:
we talked about it during 3.0 meeting -> https://youtu.be/WW5sLVHa0b8?t=466

for now looks like we agree to go the way that this PR"
- gets in a major
- readme and GitHub release describes clearly that from now on, we will release the release candidates for new spec versions directly to master and release as minor. It is going to be up to the package (direct files) users to figure how they want to approach it, if they will just take all that we give, or if they will have proper `if` to make sure new JSON Schemas for not released specs are not used in their tools

jonaslagoni:
Added a PR to prepare `next-major-spec` for releasing as v6 with the v3 schemas, see https://github.com/asyncapi/spec-json-schemas/pull/421

smoya:
> we talked about it during 3.0 meeting -> https://youtu.be/WW5sLVHa0b8?t=466
> 
> for now looks like we agree to go the way that this PR"
> 
> * gets in a major
> * readme and GitHub release describes clearly that from now on, we will release the release candidates for new spec versions directly to master and release as minor. It is going to be up to the package (direct files) users to figure how they want to approach it, if they will just take all that we give, or if they will have proper `if` to make sure new JSON Schemas for not released specs are not used in their tools

I'm sorry I'm getting into this discussion a bit late, but I want to also share.

From a point of view of the final user, and in an exercise of getting into their shoes, I understand all released schemas under both `schemas` or `schemasWithoutId` to be final and never release candidates. Also, I would expect these to be fully supported by AsyncAPI in terms of the core combo of Spec + Parser-JS ([and maybe others](https://github.com/asyncapi/parser-js/pull/526)?). 
I would never think that something found there will not be a final and "stable" version but a RC.

From the point of view of tooling, and being owner of the Parser-JS I again emphasize what @derberg mentioned about the fact tools will get those schemas and use them automatically (as Parser-JS does now since [few days ago](https://github.com/asyncapi/parser-js/pull/822). 
So I agree **a major release** will be required in order to avoid breaking tools out there that automate dependency updates. 

However, I'm still concerned about flows user might adopt regarding sticking to particular versions of the `@asyncapi/specs` package. Even though release notes contain a note about the nature of supporting a release candidate, I'm concerned about those cases where people could install and stick with version `6.0.0` that includes the pre-release version of the schemas for AsyncAPI v3 and **never** update that dependency because "everything seems to work" (but doesn't).

Even though I always advocate for release things as earlier as possible, I want to raise again the question: does releasing this earlier gives a really clear benefit VS not doing it?

smoya:
> Even though I always advocate for release things as earlier as possible, I want to raise again the question: does releasing this earlier gives a really clear benefit VS not doing it?

On the other hand, If we don't release it, tools like Parser-JS will require to point to the "RC" versions of this package if we want to release them. That moves those tools into the same situation we are having here, so IMHO should be a all-in (JSON Schemas + Parser-JS) release. If we do it in that way, I think it makes more sense releasing earlier.

fmvilas (approved):
I think we can already merge this. Other than fixing conflicts, is ready.

fmvilas (approved):

derberg (approved):

derberg:
ready to merge, sonar complains about CI building which can be later ignored anyway, not runtime related

fmvilas:
🎉 🥳

PR #224, opened 2022-05-27
feat: add specific binding schemas

jonaslagoni:
**Description**
This PR adds 2 changes to the JSON Schema documents.

1. It fixes an issue where we allow all extensions regardless of the use of `x-`, which should not be allowed. Whether we want to propagate this across all versions, not so sure? 🤔  
2. It splits the generic `bindingsObject` into the 4 specific bindings for operation, channel, message, and server.

Related to https://github.com/asyncapi/spec/issues/507
Related to https://github.com/asyncapi/bindings/issues/113

### DIFF
--- a/definitions/3.0.0/channelBindingsObject.json
+++ b/definitions/3.0.0/channelBindingsObject.json
@@ -1,6 +1,11 @@
 {
   "type": "object",
-  "additionalProperties": true,
+  "additionalProperties": false,
+  "patternProperties": {
+    "^x-[\\w\\d\\.\\-\\_]+$": {
+      "$ref": "http://asyncapi.com/definitions/3.0.0/specificationExtension.json"
+    }
+  },
   "properties": {
     "http": {},
     "ws": {},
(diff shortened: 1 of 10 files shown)


magicmatatjahu (commented):
@jonaslagoni Two things:

- > It fixes an issue where we allow all extensions regardless of the use of x-, which should not be allowed. Whether we want to propagate this across all versions, not so sure? 🤔
  
  Do you mean that bug was only in `bindingsObject`?
  
- I think that you should also update https://github.com/asyncapi/spec-json-schemas/blob/067a129541c8ac8076734d98f8dba95e80871e50/definitions/3.0.0/components.json#L71

jonaslagoni:
> * Do you mean that bug was only in `bindingsObject`?

Yes, as you can see here old vs new: https://github.com/asyncapi/spec-json-schemas/pull/224/files#diff-ecc755fd231eb52860aef0a5ea2c062296e87a0e38e1d1781d1782d47f690fe2R2-R8

We allowed the extension of any key, instead of only with specification extension `x-`.

> * I think that you should also update

Agreed, that will be next after this, and https://github.com/asyncapi/spec-json-schemas/pull/225 is merged.

fmvilas (approved):
LGTM 👍

smoya (approved):
LGTM!

Issue #236, opened 2022-06-27
[JSON Schema Store] IDEs cannot determine the version of the JSON Schema to use


smoya:
#### Describe the bug
IDEs, like VSCode, cannot determine the version of the JSON Schema file to use based on the AsyncAPI version in use. 

![2022-06-27 at 09 48 32](https://user-images.githubusercontent.com/1083296/175913728-fb6132ad-abb8-48da-ad30-ae2588bb2f41.png)

#### How to Reproduce

- Copy any example from https://github.com/asyncapi/spec/tree/master/examples
- Change the version from `2.4.0` to `2.3.0`.
- You will still notice the IDE is using `AsyncAPI (2.4.0.json)` (as in the screenshot above).

#### Expected behavior
IDEs should pick up the proper JSON Schema file based on the version selected on the AsyncAPI file.

The solution can be found in https://github.com/SchemaStore/schemastore/pull/2229#issuecomment-1167587740

PR #237, opened 2022-07-06
feat: add schema listing all Schema-Store JSON Schema docs

smoya:
**Description**

Partially fixes https://github.com/asyncapi/spec-json-schemas/issues/236, [pending a PR](https://github.com/SchemaStore/schemastore/compare/master...smoya:feat/asyncapiVersions?expand=1) on JSON Schema-Store repository.

This PR adds a new schema file that lists all JSON Schema files for all AsyncAPI spec versions. This is required for [Schema-Store](https://www.schemastore.org/) IDE plugins to support all versions.

**Related issue(s)**
https://github.com/asyncapi/spec-json-schemas/issues/236

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,4 @@
 module.exports = {
-  'all.schema-store': require('./schemas/all.schema-store.json'),
   '1.0.0': require('./schemas/1.0.0.json'),
   '1.1.0': require('./schemas/1.1.0.json'),
   '1.2.0': require('./schemas/1.2.0.json'),
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 module.exports = {
+  'all.schema-store': require('./schemas/all.schema-store.json'),
   '1.0.0': require('./schemas/1.0.0.json'),
   '1.1.0': require('./schemas/1.1.0.json'),
   '1.2.0': require('./schemas/1.2.0.json'),
--- a/schemas/all.schema-store.json
+++ b/schemas/all.schema-store.json
@@ -1,0 +1,149 @@
+{
+  "$id": "http://asyncapi.com/schema-store/all.schema-store.json",
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "title": "JSON Schema documents for all AsyncAPI spec versions",
+  "description": "All AsyncAPI JSON Schema documents listed in one file. Needed for serving all documents through schemastore.org",
+  "type": "object",
+  "oneOf": [
+    {
+      "allOf": [
+        {
+          "properties": {
+            "asyncapi": {
+              "const": "1.0.0"
+            }
+          }
+        },
+        {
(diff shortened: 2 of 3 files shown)


schemas/all.schema-store.json
```diff
@@ -0,0 +1,149 @@
...
+          "properties": {
+            "asyncapi": {
+              "const": "1.0.0"
+            }
+          }
+        },
+        {
+          "$ref": "http://asyncapi.com/schema-store/1.0.0.json"
+        }
+      ]
```
jonaslagoni, line 20: Would `if` not make more sense? 🤔 

```json
  "if": {
    "properties": {
      "asyncapi": {
        "const": "1.0.0"
      }
    }
  },
  "then": {
    "$ref": "http://asyncapi.com/schema-store/1.0.0.json"
  },
```

index.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+  'all.schema-store': require('./schemas/all.schema-store.json'),
```
jonaslagoni, line 2: Some would see this as a breaking change, or at least I think our parser does 🤔 Cause we are breaking the assumption that what is exposed here is a map of valid AsyncAPI versions.

https://github.com/asyncapi/parser-js/blob/c962a7e7278f5fc8652621d8881032fceaa79a28/lib/parser.js#L79

i.e. I wonder if the parser would accept:

```yml
asyncapi: "all.schema-store"
...
```

jonaslagoni (changes requested):

index.js
derberg, line 2: is it necessary to expose it on API level?

smoya: The reality is that it isn't. I will just need to modify the tests as they expect all schemas to be exposed.
Afaik nothing else will be needed.

schemas/all.schema-store.json
smoya: I just followed the example they provided: https://raw.githubusercontent.com/ory/hydra/v1.7.0/.schema/config.schema.json

It makes more sense like it is now to me but no strong opinion.

index.js
smoya: This is now solved by not exporting this schema anymore. See https://github.com/asyncapi/spec-json-schemas/pull/237#discussion_r915695825

schemas/all.schema-store.json
derberg: ok, just to make 100% I understand why we need it.

we solve issue with not recognized versions, this is clear

and we solve it by introducing this schema, yes, but why? does it mean that this schema will be in the end the only schema that we reference in schema store or? are we exposing it on the website too?

derberg (approved):
LGTM

@dalelane @fmvilas ?

fmvilas (approved):

dalelane (approved):

PR #239, opened 2022-07-08
feat: add binding versions

jonaslagoni:
**Description**
This PR is a suggestion on how to handle binding versions as well as integrate them into the specification.

### Binding versioning
Each binding should now be located as:
```
bindings:
  amqp:
    0.2.0:
    x.x.x:
  kafka:
    ...
```

The binding `$id`s will now contain the accurate version, so instead of `"$id": "http://asyncapi.com/bindings/amqp/channel.json"` it will have `"$id": "http://asyncapi.com/bindings/amqp/0.2.0/channel.json"`.

This setup will allow us to work on bindings alongside the spec itself, and continuously release new binding versions.

### Integrating with spec
One of the core issues with bindings is that some binding versions might only fit together with specific versions of the spec. To allow for this finetuned control, we can now reference the binding versions exactly as they are compatible and leave out versions when necessary:

```json
    "amqp": {
      "oneOf": [
        {
          "$ref": "http://asyncapi.com/bindings/amqp/0.2.0/channel.json"
        }
      ]
    },
```

### DIFF
--- a/bindings/amqp/0.2.0/channel.json
+++ b/bindings/amqp/0.2.0/channel.json
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "http://asyncapi.com/bindings/amqp/channel.json",
+  "$id": "http://asyncapi.com/bindings/amqp/0.2.0/channel.json",
   "title": "AMQP channel bindings object",
   "description": "This object contains information about the channel representation in AMQP.",
   "type": "object",
@@ -80,7 +80,6 @@
       ],
       "description": "The version of this binding. If omitted, 'latest' MUST be assumed."
     }
-
   },
   "oneOf": [
     {
--- a/bindings/amqp/0.2.0/message.json
+++ b/bindings/amqp/0.2.0/message.json
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "http://asyncapi.com/bindings/amqp/message.json",
+  "$id": "http://asyncapi.com/bindings/amqp/0.2.0/message.json",
   "title": "AMQP message bindings object",
   "description": "This object contains information about the message representation in AMQP.",
   "type": "object",
--- a/bindings/amqp/0.2.0/operation.json
+++ b/bindings/amqp/0.2.0/operation.json
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "http://asyncapi.com/bindings/amqp/operation.json",
+  "$id": "http://asyncapi.com/bindings/amqp/0.2.0/operation.json",
   "title": "AMQP operation bindings object",
   "description": "This object contains information about the operation representation in AMQP.",
   "type": "object",
--- a/definitions/3.0.0/channelBindingsObject.json
+++ b/definitions/3.0.0/channelBindingsObject.json
@@ -9,7 +9,13 @@
   "properties": {
     "http": {},
     "ws": {},
-    "amqp": {},
+    "amqp": {
+      "oneOf": [
+        {
+          "$ref": "http://asyncapi.com/bindings/amqp/0.2.0/channel.json"
+        }
+      ]
+    },
     "amqp1": {},
     "mqtt": {},
     "mqtt5": {},
--- a/definitions/3.0.0/components.json
+++ b/definitions/3.0.0/components.json
@@ -71,25 +71,25 @@
     "serverBindings": {
       "type": "object",
       "additionalProperties": {
-        "$ref": "http://asyncapi.com/definitions/3.0.0/bindingsObject.json"
+        "$ref": "http://asyncapi.com/definitions/3.0.0/serverBindingsObject.json"
       }
     },
     "channelBindings": {
       "type": "object",
       "additionalProperties": {
-        "$ref": "http://asyncapi.com/definitions/3.0.0/bindingsObject.json"
+        "$ref": "http://asyncapi.com/definitions/3.0.0/channelBindingsObject.json"
       }
     },
     "operationBindings": {
       "type": "object",
       "additionalProperties": {
-        "$ref": "http://asyncapi.com/definitions/3.0.0/bindingsObject.json"
+        "$ref": "http://asyncapi.com/definitions/3.0.0/operationBindingsObject.json"
       }
     },
     "messageBindings": {
       "type": "object",
       "additionalProperties": {
-        "$ref": "http://asyncapi.com/definitions/3.0.0/bindingsObject.json"
+        "$ref": "http://asyncapi.com/definitions/3.0.0/messageBindingsObject.json"
       }
     }
   },


jonaslagoni:
Before converting all the bindings to this exact setup, I think we need to agree on the approach beforehand. Tagging code owners and known interested parties: @fmvilas @smoya @derberg @dalelane @magicmatatjahu

fmvilas:
We considered this setup back then and ditched this idea because it allows having more than one version of a single binding, which is weird. However, since you're using `oneOf` and the version key is part of the binding, it may make more sense 🤔 Love it because it makes the tooling easier.

Referencing https://github.com/asyncapi/bindings/issues/62 because it's related.

jonaslagoni:
> With this approach, the only hard requirement is that JSON Schemas for bindings should contain `bindingVersion` property. WIthout it, none of the versions included in `oneOf` could be determined.

We default to the latest binding version if nothing is specified.

> As a side note, maybe, we could do a similar script as you @jonaslagoni did for creating a new JSON Schema file for new AsyncAPI spec version but for creating a new binding or a new version of a binding. This could go into a new issue/feature request.

We should definitely do that yes 🙂

smoya:
> We default to the latest binding version if nothing is specified.

That could potentially break documents if a new **major** version of the binding is released. Documents that were working with binding `A` without specifying the version, assuming the latest was `1.0.0` but suddenly `2.0.0` is released

jonaslagoni:
> That could potentially break documents if a new **major** version of the binding is released. Documents that were working with binding `A` without specifying the version, assuming the latest was `1.0.0` but suddenly `2.0.0` is released

Yea you are right, might need to redefine that behavior when introducing this.

dalelane (commented):
looks like a good idea to me 👍

derberg:
why it targets `next-major-spec` and not `master`? I don't see it relying on any changes related to 3.0 spec changes, or did I miss something? we can release it sooner right? just another major, or?

jonaslagoni:
> why it targets `next-major-spec` and not `master`? I don't see it relying on any changes related to 3.0 spec changes, or did I miss something? we can release it sooner right? just another major, or?

I actually did not count on integrating the bindings for version 2, which is why I targeted `next-major-spec`. 

The only reason for this is to err on caution, as we are removing a schema if we targeted an version 2, as it might already be in use.

fmvilas (dismissed):

derberg:
@jonaslagoni sorry, I mean that we do not need to wait for 3.0 as we can just release `spec-json-schemas` 4.0.0

jonaslagoni:
> @jonaslagoni sorry, I mean that we do not need to wait for 3.0 as we can just release `spec-json-schemas` 4.0.0

I know, but as i said, I did not plan to backtrack this to AsyncAPI 2.x, that said its definitely possible to do, using the same setup as here. 

Are you implying that is what er should do?

derberg:
> Are you implying that is what we should do? Or? 😃

yeah, kinda 😄 

- it is a great feature, and it would be nice if we ship it sooner to people, get feedback, adjust
- it is a breaking change. I'm in a club of people that prefer not to introduce too many changes at once, so it is easier to debug errors

but I'm definitely not planning to die on that hill if others don't share my opinion 😃

jonaslagoni:
Alright so here are the options:
1. We keep the PR as it is now, and only introduce this change for 3.0.0 (for now, we can decide to do it later if we want to). It will mean we delay when users get access to the binding schemas and it prevents a new major version.
2. We re-create this PR to target the master branch and release it as a new major version.

@fmvilas @dalelane @derberg those are your possibilities in short 😅 Let me know what you would like to do, I have no strong feelings either other then option 1 is the easiest from my point of view 😆

smoya:
I also see more value in releasing this sooner rather than waiting for 3.0. +1 to @derberg 's reasoning. It provides a lot of value.

fmvilas:
Yeah, don't have a strong opinion either but Lukasz's point makes sense.

dalelane (approved):
sooner sounds good to me too 👍

derberg:
@jonaslagoni fyi https://github.com/asyncapi/bindings/pull/143/

jonaslagoni:
Its a good question @Souvikns, I choose to do it the "easy" way that required less to do by the maintainers i.e. that you just add all the binding schemas for all the binding versions to the oneOf.

Here a validator tries all the different schemas and once it tries to validate the binding version (because it is a constant and can only be one value) it fails because the constant is not right. Does that make sense? 


I cant recall if I made sure the latest (when bindingVersion is not defined) also works now that I think about it 🤔 Might have pushed it off for future me 😄

jonaslagoni:
> @jonaslagoni fyi https://github.com/asyncapi/bindings/pull/143/

Guess I need to manually apply that one

derberg:
> Guess I need to manually apply that one

that one and probably other changes that were introduced since you opened this PR. Or just copy/paste, as I guess you did not modify original schemas

KhudaDad414 (dismissed):
this perfectly solves the problem of preserving different binding versions. thanks @jonaslagoni

jonaslagoni:
> @jonaslagoni lemme know when you wanna continue with this one upfront, so I will treat it with super high priority and help get approvals so it is quickly merged and we can remove JSON Schemas from `bindings` repo and not have the split in 2 different locations

I am swamped with other stuff at the moment, but I will focus on this either next week or the week after that.

jonaslagoni:
Alright, fixed the example document and added basic test.

We need https://github.com/asyncapi/spec-json-schemas/pull/289 merged and after that can we get this merged so I can finish this PR 😅

fmvilas:
@jonaslagoni PR #289 has been merged already. I think you have to check this PR again because it contains a bunch of changes not related to the topic.

CODEOWNERS
```diff
@@ -6,4 +6,13 @@
...
+* @fmvilas @derberg @dalelane @smoya @char0n @asyncapi-bot-eve
+
+
+/bindings/anypointmq/       @GeraldLoeffler
+/bindings/ibmmq/            @rcoppen
+/bindings/kafka/            @lbroudoux @dalelane
+/bindings/googlepubsub/     @whitlockjc
+/bindings/solace/           @damaru-inc @CameronRushton
+/bindings/*.json            @KhudaDad414
+/bindings/pulsar/           @VisualBean
```
jonaslagoni, line 18: Just a friendly reminder, you will need to invite all as contributors once it's fully merged 🙂

jonaslagoni:
I have updated the PR with the latest changes on next and added the new pulsar bindings, should be ready for final review ✌️

dalelane (approved):
👍

derberg:
just to clarify the previous discussion. So the final decision is to put it in 3.0 and not sooner?

jonaslagoni:
> So the final decision is to put it in 3.0 and not sooner?

I just don't want to spend the time splitting up the PR into one for `next` and wait for it to be released, and then add support for it in `next-major-spec` 😅 Especially considering the time this PR took.

jonaslagoni:
We can merge this and backtrack it to `next` if anybody wants to drive it 👍

jonaslagoni:
And we still dont know what to do about the specification extension, as it's currently explicitly linked to spec 3.0 in each binding version. So it's not as easy as saying "let's do it", problems will arise when we try to backport it.

derberg:
I do not have super hard feelings about it.
I know that there are people using bindings as they want (not following binding schemas entirely) on production, so releasing these before spec 3.0 will cause them some headaches. 

So yeah, let's just merged it for 3.0 only for now

derberg (approved):

CODEOWNERS
derberg: can you also ping these maintainers in https://github.com/asyncapi/bindings/issues/113 so they are aware 🙏🏼

jonaslagoni: Will do once merged 👍

jonaslagoni:
@derberg I will let you folks decide when adequate enough time has passed for reviewers to respond and for merge it.

fmvilas (approved):
LGTM 👍

smoya (approved):
🚀🌔 LGTM!

jonaslagoni:
So yea

char0n:
Great work here! :tada: 

Regarding AsyncAPI 2.x.y - I'm assuming that every binding version that is available available is allowed to be used for any AsyncAPI version right? Or do we have somewhere the matching map of which binding versions is allowed for what AsyncAPI 2.x.y spec?

jonaslagoni:
@char0n yea there is still no consensus or progress in the discussion regarding that, as the spec 3.0 changes need to be done first before figuring out what to do about the compatibility between versions of bindings: https://github.com/asyncapi/bindings/issues/182

PR #248, opened 2022-09-09
feat!: remove unsupported specs from NodeJS export

### LINKED ISSUE #238: drop asyncapi json schemas prior 2.0 and release 4.0
we still expose AsyncAPI 1.0, 1.1 and 1.2 but we do not really support it anymore.

we should remove it and release new major

these old schemas use draft 4 as base. The new ones use draft 7. So in parser we are forced to point to draft 4 as meta schema for ajv. So we do something in the parser that do not make much sense, as parser do not support asyncapi prior 2.0

thoughts?

cc approvers @fmvilas @dalelane

magicmatatjahu:
**Description**

- remove unsupported specs from NodeJS export 
- update `Readme.md`

**Related issue(s)**
Resolves #238


supported.d.ts
```diff
@@ -0,0 +1,13 @@
+import type { JSONSchema7 } from 'json-schema';
+
+declare module '@asyncapi/specs/supported' {
+  declare const schemas: {
```
smoya, line 4: Could we somehow avoid having to add every new supported version into this object?
Could we do the opposite, to maintain a list of non-supported (or regex, or whatever) and from that build this?

index.d.ts
```diff
@@ -0,0 +1,18 @@
+import type { JSONSchema7 } from 'json-schema';
+
+declare module '@asyncapi/specs' {
```
smoya, line 3: Are we gonna remove `index.js` in favor of this?

magicmatatjahu: but then we need transpilation TS -> JS process.

smoya: Yes, otherwise we manually maintain both versions.

magicmatatjahu: Can be done, lets wait for others opinion.

supported.d.ts
smoya: I'm not sure I get it. You know, still new into this world 😆 . What is the feature we lose if we do it as I suggested?

magicmatatjahu: I tested your idea and TS did not see in other packages the filtered versions without the unsupported ones, instead of `2.0.0` etc it showed me string type and not union. Maybe with TS -> it will see it.

README.md
```diff
@@ -29,14 +36,18 @@ All test cases and parsers coverage can be found [here](https://asyncapi.github.
 ## Installation
 
 ### NodeJS
+
 ```bash
 npm install @asyncapi/specs
+yarn add @asyncapi/specs
```
dalelane, line 42: I'm not super-familiar with `yarn` - are we saying that `yarn` is required to use this module? Or that it's an option?

magicmatatjahu: Yarn is another Node package manager, created by Facebook, next to the npm (default one installed by NodeJS), but a lot of people use it because it has a lot of features that npm hasn't (but latest versions have similar functionality). It has a another CLI API to install packages so yarn is a option, how to install package by this manager.

Should we describe it as:

```bash
npm install @asyncapi/specs
// OR by Yarn
yarn add @asyncapi/specs
```

wdyt?

magicmatatjahu: Done :)

supported.d.ts
magicmatatjahu: No longer current. PR has changed its "character".

index.d.ts
magicmatatjahu: No longer current. PR has changed its "character".

derberg: yeah, I definitely prefer we generate `index.js` as it doesn't make sense to manually duplicate

package-lock.json
```diff
@@ -8342,4 +8342,4 @@
       "dev": true
     }
   }
-}
```
derberg, line 8345: it is strange that you added new dependency, but this is the only change in package-lock 🤔

README.md
```diff
@@ -1,18 +1,25 @@
 ![npm](https://img.shields.io/npm/v/@asyncapi/specs?style=for-the-badge) ![npm](https://img.shields.io/npm/dt/@asyncapi/specs?style=for-the-badge)
+
 > If you are currently using version 2, check out [migration guideline to version 3](./migrations/Migrate%20to%20version%203.md). You might be able to update it without any change.
```
derberg, line 3: we need to update this part a bit. 
we talk about migration from 2 ->3 but then there is v4, so might be confusing

maybe we need dedicated section, like `Migrations`, where we list path from 2 -> 3 and explain that 3 -> 4 means that we no longer support versions prior 2.0 which means migration path is to switch to 2.x version of the spec

derberg (changes requested):
left some comments, sorry for being late here

index.d.ts
```diff
@@ -0,0 +1,14 @@
...
+  declare const specs: {
```
dalelane, line 4: should we have the `declare` here? 

when I tried opening this, I got an error:

```
A 'declare' modifier cannot be used in an already ambient context.ts(1038)
```

I've not created typescript declarations before, so I'm not really sure what the implications of this are - but removing the inner `declare` made the IDE error go away.  🤷

fmvilas:
FYI, this PR is blocking #264. [See Lukasz' comment](https://github.com/asyncapi/spec-json-schemas/pull/264#issuecomment-1273027917).

derberg:
@magicmatatjahu pingy pongy

magicmatatjahu:
@derberg Done :)

index.d.ts
magicmatatjahu: hmm, strange. I wasn't getting that error, but I removed that keyword `declare` before `const`. On the stackoverflow people write that since some version of TS the word `declare` is not needed when we declare it inside a module.

Thanks!

dalelane (approved):

index.d.ts
smoya: Are we gonna do this in this PR or in another one right after merging this one?

smoya: Due to the fact this PR blocks https://github.com/asyncapi/spec-json-schemas/pull/264 (See [here](https://github.com/asyncapi/spec-json-schemas/pull/264#issuecomment-1273027917)), I suggest we first merge this one, we then enable TS transpilation asap in another PR

smoya (approved):
Due to the fact this PR blocks https://github.com/asyncapi/spec-json-schemas/pull/264 (See https://github.com/asyncapi/spec-json-schemas/pull/264#issuecomment-1273027917), I suggest we first merge this one, we then enable TS transpilation asap in another PR

derberg (approved):
LGTM

fmvilas (approved):

fmvilas:
@magicmatatjahu wanna do the honors? 😄

magicmatatjahu:
@fmvilas Sure! :)

PR #264, opened 2022-09-28
feat!: add new channels and operations objects

fmvilas:
### Description
It adds the new `operations` and `channels` objects for version 3.0.0.

### Related issue(s)
https://github.com/asyncapi/spec/pull/827
https://github.com/asyncapi/spec/pull/806

### DIFF
--- a/definitions/3.0.0/asyncapi.json
+++ b/definitions/3.0.0/asyncapi.json
@@ -38,6 +38,9 @@
     "channels": {
       "$ref": "http://asyncapi.com/definitions/3.0.0/channels.json"
     },
+    "operations": {
+      "$ref": "http://asyncapi.com/definitions/3.0.0/operations.json"
+    },
     "components": {
       "$ref": "http://asyncapi.com/definitions/3.0.0/components.json"
     },
--- a/definitions/3.0.0/asyncapi.json
+++ b/definitions/3.0.0/asyncapi.json
@@ -5,8 +5,7 @@
   "type": "object",
   "required": [
     "asyncapi",
-    "info",
-    "channels"
+    "info"
   ],
   "additionalProperties": false,
   "patternProperties": {
--- a/definitions/3.0.0/channel.json
+++ b/definitions/3.0.0/channel.json
@@ -7,8 +7,12 @@
     }
   },
   "properties": {
-    "$ref": {
-      "$ref": "http://asyncapi.com/definitions/3.0.0/ReferenceObject.json"
+    "address": {
+      "type": ["string", "null"],
+      "description": "An optional string representation of this channel's address. The address is typically the \"topic name\", \"routing key\", \"event type\", or \"path\". When `null` or absent, it MUST be interpreted as unknown. This is useful when the address is generated dynamically at runtime or can't be known upfront. It MAY contain Channel Address Expressions."
+    },
+    "messages": {
+      "$ref": "http://asyncapi.com/definitions/3.0.0/channelMessages.json"
     },
     "parameters": {
       "type": "object",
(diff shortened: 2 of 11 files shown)


definitions/3.0.0/channelMessages.json
```diff
@@ -0,0 +1,9 @@
+{
```
derberg, line 1: I understand we need this because messages are not by default list

if I got it right, just ignore, you just got `rubber-ducked`

schemas/3.0.0.json
```diff
@@ -307,11 +307,11 @@
                 "minLength": 1
```
derberg, line 307: this one is generated afaik, no edits

definitions/3.0.0/asyncapi.json
```diff
@@ -5,8 +5,7 @@
   "type": "object",
   "required": [
     "asyncapi",
-    "info",
-    "channels"
+    "info"
```
derberg, line 8: you need to add `operations` here 

I recommend you add some online JSON vs JSON Schema validator, 
You can bundle all these into one schema with `npm run build` and then manually check against one of the new examples.

derberg (changes requested):
Left few comments

definitions/3.0.0/channel.json
```diff
@@ -7,8 +7,12 @@
     }
   },
   "properties": {
-    "$ref": {
-      "$ref": "http://asyncapi.com/definitions/3.0.0/ReferenceObject.json"
```
smoya, line 11: For the reference, this change will solve https://github.com/asyncapi/spec/issues/699 if I'm correct, please confirm @char0n

char0n: @smoya yes I can confirm.

definitions/3.0.0/operationTrait.json
```diff
@@ -1,41 +1,4 @@
...
-      "items": {
-        "$ref": "http://asyncapi.com/definitions/3.0.0/SecurityRequirement.json"
-      }
-    },
-    "bindings": {
-      "$ref": "http://asyncapi.com/definitions/3.0.0/operationBindingsObject.json"
-    }
-  },
-  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$ref": "http://asyncapi.com/definitions/3.0.0/operation.json",
```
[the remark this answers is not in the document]
smoya: I'm confused. What is the intended effect here?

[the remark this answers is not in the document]
fmvilas: Now operationTrait and operation objects are identical. I'm trying to reuse operation but assigning a different ID (operaitonTrait).

definitions/3.0.0/asyncapi.json
fmvilas: But `operations` is not required 🤔

definitions/3.0.0/operationTrait.json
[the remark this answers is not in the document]
fmvilas: I just tested it and it works 👍 Seems possible after all.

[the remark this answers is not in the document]
fmvilas: In the end, I'm not using it. Otherwise, we could use traits inside traits inside traits 🤯 Let's keep it simple 😄

fmvilas:
Alright, tested it on https://www.jsonschemavalidator.net/ and seems to work 👍

smoya:
> Alright, tested it on https://www.jsonschemavalidator.net/ and seems to work 👍

How cool would be to have this in place! https://github.com/asyncapi/studio/issues/207

definitions/3.0.0/channelMessages.json
derberg: yup, thanks!

definitions/3.0.0/asyncapi.json
derberg: oh yeah :)

derberg (approved):

fmvilas:
@derberg I'm confused now. Should the title say `feat!` or `feat`? 🤔 

Ping @dalelane

dalelane (approved):

derberg:
Must be `!` because last time there was no and `[v3.2.0-next-major-spec.1](https://github.com/asyncapi/spec-json-schemas/releases/tag/v3.2.0-next-major-spec.1)` was released instead of 4.0 RC

I only suggest that maybe we merge https://github.com/asyncapi/spec-json-schemas/pull/248 first, and release v4, and then this PR so we have v5 RC

derberg:
@fmvilas 4.0 is released, I think we can release this one.

since there is 3.2 rc1, I'm not 100% sure if we get 5.0 rc1, but from logical point of view, we should

fmvilas:
Alright, let's try 🚀 😄

PR #271, opened 2022-10-04
chore: add Sergio and Vladimir as code owners

fmvilas:
### Related issues

https://github.com/asyncapi/spec/issues/844
https://github.com/asyncapi/spec/issues/845


fmvilas:
Can I get an amen? @derberg

PR #288, opened 2022-11-21
fix: generated index.d.ts file using tsc command (#287)

### LINKED ISSUE #287: Cannot build index.d.ts file due to an invalid export syntax
#### Describe the bug
Since the option `esModuleInterop: true` is not set to the project, the file index.d.ts cannot be compiled. This happens because the module definition of the file uses an invalid export syntax, when it should use the `export=` syntax type. This is described better on this documentation: [Module.d.ts](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html)

```
Note that using export default in your .d.ts files requires [esModuleInterop: true](https://www.typescriptlang.org/tsconfig#esModuleInterop) to work. If you can’t have esModuleInterop: true in your project, such as when you’re submitting a PR to Definitely Typed, you’ll have to use the export= syntax instead
```

File path: https://github.com/asyncapi/spec-json-schemas/blob/master/index.d.ts

#### How to Reproduce
`npm run build:prod` (tested in local environment only)

Link to GitHub repository with project that has issues:
https://github.com/asyncapi/server-api

#### Expected behavior
After running the command, it should finish without any build errors.

vinibiavatti1:
**Description**
Used tsc command to recreate index.d.ts (definiton file) for index.js since it could not be compiled. The reason is that the index.js file is a commonjs file, and the definition file does not allow module augmentation with exports when commonjs is used to the main module.

**Related issue(s)**
Fixes #287

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,14 +1,11 @@
 import type { JSONSchema7 } from 'json-schema';
 
-declare module '@asyncapi/specs' {
-  const specs: {
-    '2.0.0': JSONSchema7,
-    '2.1.0': JSONSchema7,
-    '2.2.0': JSONSchema7,
-    '2.3.0': JSONSchema7,
-    '2.4.0': JSONSchema7,
-    '2.5.0': JSONSchema7,
-  }
-
-  export default specs;
-}
+declare const _exports: {
+    '2.0.0': JSONSchema7;
+    '2.1.0': JSONSchema7;
+    '2.2.0': JSONSchema7;
+    '2.3.0': JSONSchema7;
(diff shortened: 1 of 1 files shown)


derberg:
@magicmatatjahu can you 🙏🏼 have a look

magicmatatjahu (approved):
I tested new typings in the Studio as dependency and it works. Thanks for contribution!

derberg:
@magicmatatjahu thanks mate 🙏🏼 
@vinibiavatti1 thanks for contribution

/rtm

derberg (approved):

Issue #296, opened 2022-12-12
Cannot define channel's parameter as reference


magicmatatjahu:
#### Describe the bug

As we can see in the Studio:

![image](https://user-images.githubusercontent.com/20404945/207093885-14e0fc81-4a33-455d-b952-d4ae9c20d426.png)

![image](https://user-images.githubusercontent.com/20404945/207094908-f1070798-da69-4198-a557-345d1ff4a8bc.png)

In the channel's parameters we cannot make references.

#### How to Reproduce

Open https://studio.asyncapi.com/?url=https://raw.githubusercontent.com/asyncapi/spec/master/examples/streetlights-kafka.yml and convert to the JSON - right corner of the editor -> `Convert to JSON` and you should see yellow marker under `$ref` used in any channel's parameter (in YAML format you won't see error due to the https://github.com/asyncapi/studio/issues/509 issue).

#### Expected behavior

As a user I can make reference to the channel's parameter without "referencing" errors.

magicmatatjahu:
@derberg 

> I don't get why it shows up without throwing errors that AsyncAPI document doesn't follow the schema.

Seriously you can't handle if I'm missing 😅 Because we don't validate schema before resolving refs, but only after, that's you only see "warnings", not "errors".

> Also, I don't mind 👇🏼 change but honestly see no difference, other than "it looks cleaner"

Check this change: https://github.com/asyncapi/spec-json-schemas/pull/320/files#diff-b94db5fcbf97dcbdb295076705f2ad0e55939493cdf43effbea7ceff9fe5404cR4

I changed it to the `parameters` schema where we have union of ref object and parameter item.

derberg:
> Seriously you can't handle if I'm missing

do you remember my comment about your english from last week 🤣 (super internal joke 😄  )

> Because we don't validate schema before resolving refs, but only after, that's you only see "warnings", not "errors".

ups, forgot 😆 

> Check this change: https://github.com/asyncapi/spec-json-schemas/pull/320/files#diff-b94db5fcbf97dcbdb295076705f2ad0e55939493cdf43effbea7ceff9fe5404cR4
I changed it to the parameters schema where we have union of ref object and parameter item.

I'm showing you `2.5` and you point me to `2.4`.
For 2.5 and 2.6 we already had that done, and `parameters` was union already https://asyncapi.com/definitions/2.5.0/parameters.json

I'm trying to understand the change and the difference.

magicmatatjahu:
[...]

Issue #247, opened 2022-08-31
Vscode unable to resolve parts of the schema


TomTardigradeSEL:
#### Describe the bug

According to https://github.com/redhat-developer/yaml-language-server/issues/585 and https://github.com/microsoft/vscode-json-languageservice/issues/123 the use of hashes to reference eternal schemas is not supported. 
`$ref '/definitions/schemaArray' in 'http://asyncapi.com/schema-store/2.4.0.json' can not be resolved.`

#### How to Reproduce
Create a new asyncapi.json or asyncapi.yaml file in vscode with the [yaml extention](https://github.com/redhat-developer/vscode-yaml) installed. 

![image](https://user-images.githubusercontent.com/101903840/187714885-a3f54fb3-bbdf-48da-9ead-00d10cc37ce7.png)

#### Expected behavior
The schema should resolve by the language servers used by the extention.

jonaslagoni:
Without looking into it, I am pretty sure it's because the reference is not accurately resolved to this schema: https://github.com/asyncapi/spec-json-schemas/blob/83c03f222220f18ec3a54b913c4023766149bcac/schemas/2.4.0.json#L486

When this reference is encountered: https://github.com/asyncapi/spec-json-schemas/blob/83c03f222220f18ec3a54b913c4023766149bcac/schemas/2.4.0.json#L601 it MUST use the schema https://github.com/asyncapi/spec-json-schemas/blob/83c03f222220f18ec3a54b913c4023766149bcac/schemas/2.4.0.json#L483 to resolve the relevant references and NOT expect that the root `2.1.0.yaml` schema contains a schema at `#/definitions/schemaArray` i.e. the fully resolved reference should be `http://json-schema.org/draft-07/schema#/definitions/schemaArray`.

jonaslagoni:
> @jonaslagoni we know that there is a problem, but it is outside our reach, thus my suggestion is that we basically resolve all references, inline them in the schema that we publish (ignoring circular if we have these)

It's definitely a possibility, but ignoring circular is just not an option I think.

An alternative approach is that we create our "own" bundling behavior and change the `$ref`s to be absolute from the perspective of the root object I.e. the bundling behavior would change the reference https://github.com/asyncapi/spec-json-schemas/blob/83c03f222220f18ec3a54b913c4023766149bcac/schemas/2.4.0.json#L601

to `/definitions/http://json-schema.org/draft-07/schema/definitions/schemaArray`.

[...]

PR #298, opened 2022-12-13
feat: modify bundling to create refs pointing to local definitions

derberg:
Fixes https://github.com/asyncapi/spec-json-schemas/issues/247
Fixes https://github.com/asyncapi/spec-json-schemas/issues/296

This is a workaround implementation. I'm not proud of it but it is better than the current situation.

Our AsyncAPI JSON Schema is created following JSON Schema best practices and also requirements of https://github.com/hyperjump-io/json-schema-bundle - which means we have beautiful `$id` and our references are based on these, which makes schema unreadable for humans (like me that completely do not understand `$id` and `$ref` concept, but my issues could be ignored and I could learn) and are not understood by some tools (like https://github.com/redhat-developer/vscode-yaml that VSCode YAML autocompletion and validation is based on, and this is a problem for us).

- Why workaround and not something we can solve at above-mentioned bundler level? -> https://github.com/hyperjump-io/json-schema-bundle/issues/9
- Why not solve this with the above-mentioned YAML plugin managed by RedHat? -> let me answer with question: what assurance do we get it will not break in the future with further contributions to the plugin, or that other tools do it the proper way?

---
- This PR also contains generated schemas, under `schemas` dir for reference so you see what has changed in these because of the new bundler
- I tested newly generated schemas with asyncapi file on local,  and autocompletion and validation works + this way I discovered issue with `parameters`

PS. @jonaslagoni not sure if `$id` should be removed from these schemas 🤔

### DIFF
--- a/definitions/2.5.0/channelItem.json
+++ b/definitions/2.5.0/channelItem.json
@@ -13,7 +13,7 @@
     "parameters": {
       "type": "object",
       "additionalProperties": {
-        "$ref": "http://asyncapi.com/definitions/2.5.0/parameter.json"
+        "$ref": "http://asyncapi.com/definitions/2.5.0/parameters.json"
       }
     },
     "description": {
--- a/package.json
+++ b/package.json
@@ -38,13 +38,13 @@
   },
   "homepage": "https://github.com/asyncapi/spec-json-schemas#readme",
   "devDependencies": {
-    "mocha": "^10.0.0",
-    "nyc": "^15.1.0",
     "@semantic-release/commit-analyzer": "^8.0.1",
     "@semantic-release/github": "7.2.3",
     "@semantic-release/npm": "^7.0.3",
     "@semantic-release/release-notes-generator": "^9.0.1",
     "conventional-changelog-conventionalcommits": "^4.2.3",
+    "mocha": "^10.0.0",
+    "nyc": "^15.1.0",
     "semantic-release": "19.0.3"
   },
   "release": {
@@ -85,6 +85,7 @@
     ]
   },
   "dependencies": {
-    "@types/json-schema": "^7.0.11"
+    "@types/json-schema": "^7.0.11",
+    "json-schema-traverse": "^1.0.0"
   }
 }
--- a/tools/bundler/index.js
+++ b/tools/bundler/index.js
@@ -1,7 +1,9 @@
 const path = require('path');
 const fs = require('fs');
+const traverse = require('json-schema-traverse');
 const definitionsDirectory = path.resolve(__dirname, '../../definitions');
 const outputDirectory = path.resolve(__dirname, '../../schemas');
+const JSON_SCHEMA_PROP_NAME = 'json-schema-draft-07-schema';
 console.log(`Looking for separate definitions in the following directory: ${definitionsDirectory}`);
 console.log(`Using the following output directory: ${outputDirectory}`);
 
@@ -29,14 +31,68 @@ console.log(`Using the following output directory: ${outputDirectory}`);
       const filePathToBundle = `file://${versionDir}/asyncapi.json`;
       const fileToBundle = await Bundler.get(filePathToBundle);
       const bundledSchema = await Bundler.bundle(fileToBundle);
-      bundledSchema.description = `!!Auto generated!! \n Do not manually edit. ${bundledSchema.description ?? ''}`;
+      const schemaWithoutURLs = modifyRefsandDefinitions(bundledSchema);
+      schemaWithoutURLs.description = `!!Auto generated!! \n Do not manually edit. ${schemaWithoutURLs.description ?? ''}`;
       const outputFile = path.resolve(outputDirectory, `${version}.json`);
       console.log(`Writing the bundled file to: ${outputFile}`);
-      await fs.promises.writeFile(outputFile, JSON.stringify(bundledSchema, null, 4));
+      await fs.promises.writeFile(outputFile, JSON.stringify(schemaWithoutURLs, null, 4));
     }catch(e) {
-      console.log(e);
+      throw new Error(e);
     }
   }
   console.log('done');
 })();
 
+
+/**
+ * we first update definitions from URL to normal names
+ * than update refs to point to new definitions, always inline never remote
+ */
+function modifyRefsandDefinitions(bundledSchema) {
+  const schemaWithoutUrls = bundledSchema;
+
+  //first we need to improve names of the definitions from URL to their names
+  for (const def of Object.keys(schemaWithoutUrls.definitions)) {
+    const newDefName = getDefinitionName(def);
+    
+    //creating copy of definition under new name so later definition stored under URL name can be removed
+    schemaWithoutUrls.definitions[newDefName] = schemaWithoutUrls.definitions[def];
+    delete schemaWithoutUrls.definitions[def]
+  }
+
+  traverse(schemaWithoutUrls, replaceRef);
+
+  return schemaWithoutUrls;
+}
+
+/**
+ * by default schemas definition names are urls like http://asyncapi.com/definitions/2.4.0/parameters.json
+ * we need to get rid of URLs and use the last fragment as new definition name like `parameters`
+ */
+function getDefinitionName(def) {
+  if (def.startsWith('http://json-schema.org')) return JSON_SCHEMA_PROP_NAME;
+
+  if (path.extname(def) !== '.json') throw `Oryginal $id values should point to JSON files. There is probably an error in one of source definitions containing definition: ${def}`;
+  return path.basename(def, '.json')
+}
+
+/**
+ * this is a callback used when traversing through json schema
+ * it is triggered with every new element of json schema
+ */
+function replaceRef(schema) {
+
+  //traversing shoudl take place only in case of schemas with refs
+  if (schema.$ref === undefined ) return;
+
+  // '#/definitions' refs are always those related to JSON Schema draft, so we just need to update them to point to json schema draft that is inlined inside schema
+  if (schema.$ref.startsWith('#/definitions')) {
+    schema.$ref = schema.$ref.replace('#/definitions/', `#/definitions/${JSON_SCHEMA_PROP_NAME}/definitions/`);
+  // '#' refs need to be updated to point to the root of inlined json schema draft
+  } else if (schema.$ref === '#') {
+    schema.$ref = `#/definitions/${JSON_SCHEMA_PROP_NAME}`;
+  // the rest of refs are those related to remote URL refst that need to be update and point to inlined versions
+  } else {
+    schema.$ref = `#/definitions/${getDefinitionName(schema.$ref)}`;
+  }
+}
--- a/tools/bundler/index.js
+++ b/tools/bundler/index.js
@@ -31,11 +31,11 @@ console.log(`Using the following output directory: ${outputDirectory}`);
       const filePathToBundle = `file://${versionDir}/asyncapi.json`;
       const fileToBundle = await Bundler.get(filePathToBundle);
       const bundledSchema = await Bundler.bundle(fileToBundle);
-      const schemaWithoutURLs = modifyRefsandDefinitions(bundledSchema);
-      schemaWithoutURLs.description = `!!Auto generated!! \n Do not manually edit. ${schemaWithoutURLs.description ?? ''}`;
+      modifyRefsandDefinitions(bundledSchema);
+      bundledSchema.description = `!!Auto generated!! \n Do not manually edit. ${bundledSchema.description ?? ''}`;
       const outputFile = path.resolve(outputDirectory, `${version}.json`);
       console.log(`Writing the bundled file to: ${outputFile}`);
-      await fs.promises.writeFile(outputFile, JSON.stringify(schemaWithoutURLs, null, 4));
+      await fs.promises.writeFile(outputFile, JSON.stringify(bundledSchema, null, 4));
     }catch(e) {
       throw new Error(e);
     }
@@ -49,20 +49,17 @@ console.log(`Using the following output directory: ${outputDirectory}`);
  * than update refs to point to new definitions, always inline never remote
  */
 function modifyRefsandDefinitions(bundledSchema) {
-  const schemaWithoutUrls = bundledSchema;
 
   //first we need to improve names of the definitions from URL to their names
-  for (const def of Object.keys(schemaWithoutUrls.definitions)) {
+  for (const def of Object.keys(bundledSchema.definitions)) {
     const newDefName = getDefinitionName(def);
     
     //creating copy of definition under new name so later definition stored under URL name can be removed
-    schemaWithoutUrls.definitions[newDefName] = schemaWithoutUrls.definitions[def];
-    delete schemaWithoutUrls.definitions[def]
+    bundledSchema.definitions[newDefName] = bundledSchema.definitions[def];
+    delete bundledSchema.definitions[def]
   }
 
-  traverse(schemaWithoutUrls, replaceRef);
-
-  return schemaWithoutUrls;
+  traverse(bundledSchema, replaceRef);
 }
 
 /**
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,11245 @@
 {
   "name": "@asyncapi/specs",
   "version": "4.0.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "@asyncapi/specs",
+      "version": "4.0.1",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@types/json-schema": "^7.0.11",
+        "json-schema-traverse": "^1.0.0"
+      },
+      "devDependencies": {
+        "@semantic-release/commit-analyzer": "^8.0.1",
+        "@semantic-release/github": "7.2.3",
+        "@semantic-release/npm": "^7.0.3",
+        "@semantic-release/release-notes-generator": "^9.0.1",
+        "conventional-changelog-conventionalcommits": "^4.2.3",
+        "mocha": "^10.0.0",
+        "nyc": "^15.1.0",
+        "semantic-release": "19.0.3"
+      }
+    },
+    "node_modules/@babel/code-frame": {
+      "version": "7.8.3",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
+      "integrity": "sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g==",
+      "dev": true,
+      "dependencies": {
+        "@babel/highlight": "^7.8.3"
+      }
+    },
+    "node_modules/@babel/core": {
+      "version": "7.12.13",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.12.13.tgz",
+      "integrity": "sha512-BQKE9kXkPlXHPeqissfxo0lySWJcYdEP0hdtJOH/iJfDdhOCcgtNCjftCJg3qqauB4h+lz2N6ixM++b9DN1Tcw==",
+      "dev": true,
+      "dependencies": {
+        "@babel/code-frame": "^7.12.13",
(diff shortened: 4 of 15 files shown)


definitions/2.5.0/channelItem.json
```diff
@@ -13,7 +13,7 @@
     "parameters": {
       "type": "object",
       "additionalProperties": {
-        "$ref": "http://asyncapi.com/definitions/2.5.0/parameter.json"
+        "$ref": "http://asyncapi.com/definitions/2.5.0/parameters.json"
```
derberg, line 16: this is fix for https://github.com/asyncapi/spec-json-schemas/issues/296

tools/bundler/index.js
```diff
@@ -29,14 +31,65 @@ console.log(`Using the following output directory: ${outputDirectory}`);
       const filePathToBundle = `file://${versionDir}/asyncapi.json`;
       const fileToBundle = await Bundler.get(filePathToBundle);
       const bundledSchema = await Bundler.bundle(fileToBundle);
+      modifyRefsandDefinitions(bundledSchema);
```
fmvilas, line 34: It took me a few seconds to understand what it meant. I was thinking: "what the hell is a Refsand Definition now?" 😂 

```suggestion
      modifyRefsAndDefinitions(bundledSchema);
```

tools/bundler/index.js
```diff
@@ -29,14 +31,65 @@ console.log(`Using the following output directory: ${outputDirectory}`);
...
   }
   console.log('done');
 })();
 
+
+/**
+ * we first update definitions from URL to normal names
+ * than update refs to point to new definitions, always inline never remote
+ */
+function modifyRefsandDefinitions(bundledSchema) {
```
fmvilas, line 51:
```suggestion
function modifyRefsAndDefinitions(bundledSchema) {
```

tools/bundler/index.js
```diff
@@ -29,14 +31,65 @@ console.log(`Using the following output directory: ${outputDirectory}`);
...
+}
+
+/**
+ * by default schemas definition names are urls like http://asyncapi.com/definitions/2.4.0/parameters.json
+ * we need to get rid of URLs and use the last fragment as new definition name like `parameters`
+ */
+function getDefinitionName(def) {
+  if (def.startsWith('http://json-schema.org')) return JSON_SCHEMA_PROP_NAME;
+
+  if (path.extname(def) !== '.json') throw `Oryginal $id values should point to JSON files. There is probably an error in one of source definitions containing definition: ${def}`;
```
fmvilas, line 72:
```suggestion
  if (path.extname(def) !== '.json') throw `Original $id values should point to JSON files. There is probably an error in one of the source definitions containing definition: ${def}`;
```

fmvilas (changes requested):
Just small suggestions.

jonaslagoni (approved):
LGTM 👍 

So my thoughts the current `$id` and `$ref` situation is that we have solved the issue in the bundled schemas now, which is perfect 👌 

If however, people want to use a single schema, instead of the bundled one, we still support dynamically fetching those schemas remotely. Which I technically think is good, whether there is a use-case for it is another thing 😆 Cause I don't see it, just from a technical standpoint.

So if we don't see that as a "thing", we could remove the `$id` yea.

package.json
```diff
@@ -85,6 +85,7 @@
     ]
   },
   "dependencies": {
-    "@types/json-schema": "^7.0.11"
+    "@types/json-schema": "^7.0.11",
+    "json-schema-traverse": "^1.0.0"
```
magicmatatjahu, line 89: It should be installed as dev dependency.

tools/bundler/index.js
```diff
@@ -29,14 +31,65 @@ console.log(`Using the following output directory: ${outputDirectory}`);
...
+
+  traverse(bundledSchema, replaceRef);
+}
+
+/**
+ * by default schemas definition names are urls like http://asyncapi.com/definitions/2.4.0/parameters.json
+ * we need to get rid of URLs and use the last fragment as new definition name like `parameters`
+ */
+function getDefinitionName(def) {
+  if (def.startsWith('http://json-schema.org')) return JSON_SCHEMA_PROP_NAME;
```
magicmatatjahu, line 70: I know that we use only draft-07, however we should be hardcoded here only form draft-07 schema, so please change it to the:

```suggestion
  if (def.startsWith('http://json-schema.org/draft-07')) return JSON_SCHEMA_PROP_NAME;
```

magicmatatjahu (commented):
Two suggestions :)

magicmatatjahu:
> So if we don't see that as a "thing", we could remove the $id yea, but I would probably wait and see tbh 🙂

I would leave it at that, even if we currently have no use case for that.

package.json
derberg: ups, actually, it should not be there but in the `package.json` from `tools/bundler`. Thanks good catch!

tools/bundler/index.js
derberg: I accepted it initially but then reverted. It adds no value but causes complexity as in older spec versions we used draft 4. This would complicate code if I would like to have clear info about versions there

derberg:
@magicmatatjahu @jonaslagoni the problem is that once I leave `$id` then parser will not work with my changes because of errors like:
```
{
message: "can't resolve reference #/definitions/contact from…d http://asyncapi.com/definitions/2.5.0/info.json", 
missingRef: 'http://asyncapi.com/definitions/2.5.0/info.json#/definitions/contact', 
missingSchema: 'http://asyncapi.com/definitions/2.5.0/info.json'
}
```

because `ajv`, when it sees `$id`, wants to resolve $refs against it.

> If however, people want to use a single schema, instead of the bundled one, we still support dynamically fetching those schemas remotely

what do you mean by single schema? like https://github.com/asyncapi/spec-json-schemas/blob/master/definitions/2.5.0/channels.json ? the $ref that we use there is absolute so I guess `$id` is not necessary. Unless you meant something different

omg I don't think I will ever like `$id`

jonaslagoni:
@derberg yea you're right, forgot that the `$id` changes where relative references are looking. For the bundled schemas you need to remove them 🙂 

Thought you meant to remove them from the individual definitions.

tools/bundler/index.js
magicmatatjahu: Ok, thanks, I forgot about older versions.

derberg:
> Thought you meant to remove them from the individual definitions.

Actually, why not? I seriously do not get why we need it there 😆 
example:
```json
{
  "type": "object",
  "propertyNames": {
    "type": "string",
    "format": "uri-template",
    "minLength": 1
  },
  "additionalProperties": {
    "$ref": "http://asyncapi.com/definitions/2.5.0/channelItem.json"
  },
  "$id": "http://asyncapi.com/definitions/2.5.0/channels.json"
}
```

if `$ref` is anyway an absolut URL, what is `$id` for....asking for a friend 🙏🏼

magicmatatjahu:
If the schema is single, the `$id` is not needed, but if it is standalone, it is better to leave it as it determines the id of this single schema.

derberg:
@char0n thanks a lot for details explanation

> Example: JSON Schema can be physically served from different HTTP URLs, but if it contains the same $id it's considered to be the same JSON Schema. You want to bundle it just once.

now looking at my example above, reference points to http://asyncapi.com/definitions/2.5.0/channelItem.json where `$id` is different that the `$id` of schema that there reference come from

...

ok, I leave `$id` is source files and remove from bundled schema

jonaslagoni:
@derberg any update here, or are you just waiting for reviews? 🙂

derberg:
ok, so now `$id` is removed from resulting schema and autocompletion works like a charm when I check manually:

Screenshot 2023-02-09 at 18 17 31
Screenshot 2023-02-09 at 18 17 46

but in Parser it fails with:
```
Error: can't resolve reference http://asyncapi.com/definitions/2.6.0/schema.json from id #
```

but....but there is no `http://asyncapi.com/definitions/2.6.0/schema.json` reference 😭 

I thought I need `"$id": "http://asyncapi.com/schema-store/2.6.0.json"` id in schema root, but that was not a lucky guess

I ran out of ideas, need a break

jonaslagoni:
@derberg this is the parser expecting that specific schema to be present: https://github.com/asyncapi/parser-js/blob/060c1044865c7dfc2b5c688e46d11be2a3617c6f/src/schema-parser/asyncapi-schema-parser.ts#L82

If you want to avoid making a breaking change to the schema, I suggest you keep the current `$id` as keys for the definitions 🙂 I.e. instead of your current:

```
definitions:
  schema:
```

You make it:

```
definitions:
  http://asyncapi.com/definitions/2.6.0/schema.json:
```

Otherwise, this PR will be `feat!` 🙂

jonaslagoni:
After some thought, in its current form, this is actually a breaking change regardless 😆 

For tools like Studio, definitions are often split up to provide the editor with individual `$id`/`uri` and schema. This would break with the current bundled schema, as the reference require the full schema to work. I.e. if you split out `definitions.schema` a reference such as `#/definitions/schema` is no longer accessible.

I know how to change this into something that doesn't become a breaking change but not sure if you want to pursue that, let me know 🙂

derberg:
@jonaslagoni thanks, I completely forgot about AsyncAPI Schema parser 😅 

I'm not sure what happens in Studio and how come autocompletion works without this PR. I guess there is some workaround, and the schema store is not used or maybe cannot be used.

I definitely do not mind if we do it as a major release. Thanks for pointing that out.

Parser changes that will be needed once we merge this PR: https://github.com/asyncapi/parser-js/pull/726

magicmatatjahu:
@derberg @jonaslagoni Could we make (of course not in this issue) the current approach and publish two versions of "bundled" spec? With `$id`s reference mechanism and with normal (😆) references?

> For tools like Studio, definitions are often split up to provide the editor with individual $id/uri and schema: https://github.com/asyncapi/studio/blob/ef3cbafa64c06aac09d1f262275a098cc13eb76d/src/services/monaco.service.ts#L121
>
> I'm not sure what happens in Studio and how come autocompletion works without this PR. I guess there is some workaround, and the schema store is not used or maybe cannot be used.

Don't worry. I had to change logic in Studio to such a magic code, due to fact that monaco-editor doesn't recognize the references by `$id`. I mean, it recognizes but first it tries to check if another schema in internal registry has given `$id` (not this same schema, another one passed in config), then tries to download it from the web but in any way it doesn't look in the components defined in `definitions` (it looks only when relative refs are used).

Check old code when we have relative references in bundled schema https://github.com/magicmatatjahu/studio/blob/e659201e1b816d4e5d04bd96144b19aee7c1b8df/src/services/monaco.service.ts#L49 We only pass one schema (two schemas tbh, second one is `draft-07` which is on the `http` endpoint and browser blocks fetching due to fact that it's not secure connection 😆). Any questions?

jonaslagoni:
>@derberg @jonaslagoni Could we make (of course not in this issue) the current approach and publish two versions of "bundled" spec? With $ids reference mechanism and with normal (😆), relative references?

Honestly, I am not entirely against this idea. Best of both worlds and we don't have to do a breaking change.

derberg:
> @derberg @jonaslagoni Could we make (of course not in this issue) the current approach and publish two versions of "bundled" spec? With $ids reference mechanism and with normal (😆), relative references?

yeah, it was also crossing my mind few times. I think it is a way to go, but I think it must be done in this issue as for now it is a breaking change.

And I don't think it is just to avoid breaking change. When I was thinking about it, I was not currently concerned that schema file would be broken but more concerned that I make a fix for one use case, but for others, I remove some other important functionality ($id) that they highly depend on.

I'll update the PR:
- `2.6.0.json` files will stay as they are
- `2.6.0-without-$id.json` will be the ones I need for schemaStore (unless you have a better name)
- we will also have `all.schema-store-without-$id.json` with list of new schemas
- I will update readme and contribution flow
- I will close PR I did in the Parser as it is not needed at all anymore 

@magicmatatjahu THANKS!!!

I'll leave it for few days for others to react if they have some valid concerns

derberg:
@jonaslagoni I made needed changes, just please have a look if you agree that new schemas do not have to be in JS/Go packages. I do not what to expose these there if there is no use case. But if you need them on a package level, then it means we have a use case.

I won't trigger reviewers for final review until you confirm

derberg:
@jonaslagoni ok, now schemas are also exposed on the library level. Please check it works for you and then I will ping reviewers

mhanberg:
adding this to the top of my file seems to work

`# yaml-language-server: $schema=https://raw.githubusercontent.com/asyncapi/spec-json-schemas/c03428dcb043156d254f2387183bb71bacdb1fa3/schemas/2.6.0-without-%24id.json`

derberg:
@jonaslagoni should we wait for you before you check? not sure

jonaslagoni (approved):
LGTM 👍

README.md
```diff
@@ -6,10 +6,10 @@ This is a mono repository, which provides all the JSON Schema documents for vali
 
 ## Overview
 
-* This repository contains [JSON Schema](https://json-schema.org) files for all the versions of AsyncAPI specification.
+* This repository contains [JSON Schema](https://json-schema.org) files for all the versions of AsyncAPI specification. There are two types of JSON Schema files, with and without **$id** feature
```
smoya, line 9: nitpick: I would suggest we add a clarification on why those new schemas (without $id` are meant to).

index.d.ts
```diff
@@ -2,11 +2,18 @@ import type { JSONSchema7 } from 'json-schema';
...
     '2.2.0': JSONSchema7;
+    '2.2.0-without-$id': JSONSchema7;
     '2.3.0': JSONSchema7;
+    '2.3.0-without-$id': JSONSchema7;
     '2.4.0': JSONSchema7;
+    '2.4.0-without-$id': JSONSchema7;
     '2.5.0': JSONSchema7;
+    '2.5.0-without-$id': JSONSchema7;
     '2.6.0': JSONSchema7;
+    '2.6.0-without-$id': JSONSchema7;
```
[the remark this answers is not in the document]
derberg: you mean that one can assume that there is just one schema per version basically, right? but isn't `-without-$id` suffix enough? do you have some suggestions? or you basically expect we do major release?

README.md
```diff
@@ -6,10 +6,10 @@ This is a mono repository, which provides all the JSON Schema documents for vali
 
 ## Overview
 
-* This repository contains [JSON Schema](https://json-schema.org) files for all the versions of AsyncAPI specification.
+* This repository contains [JSON Schema](https://json-schema.org) files for all the versions of AsyncAPI specification. There are two types of JSON Schema files, with and without **$id** feature. We need two versions of schemas because of the differences it tooling implementation of JSON Schema `$ref` and `$id` keywords. Some implementations treat `$id` by default as prefix reference for `$ref` and require it, therefore it is needed to properly correlate `$ref` and `$id` values. Unfortunately other tools do not understand `$id` values and fail dereferencing. This is why we need two different versions of schemas, with and without the `$id`.
```
smoya, line 9: ❤️

index.d.ts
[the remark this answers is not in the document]
smoya: As this is just a workaround, can we just generate them but not export them? Only include it in the `all.schema-store.json`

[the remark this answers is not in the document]
derberg: yeah, I did not expose it in first iteration as I do not need it, but @jonaslagoni needs it in modelina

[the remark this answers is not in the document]
smoya: This change, once ported to `next-major` branch, will make https://github.com/asyncapi/parser-js/blob/next-major/src/schema-parser/asyncapi-schema-parser.ts#L38-L53 to provide wrong types.

TBH is not an issue with this PR per se. The way we are exporting the schemas is not helping. I think we could introduce the following separation in `next-major` branch:

```ts
declare const _exports: {
    'schemas': {
        // ...
    },
    'schemas-without-$id': {
        // ...
    }
}
```

Do you think it makes sense? If it does, +1 to merge this and start doing those changes in `next-major`.

cc @magicmatatjahu @jonaslagoni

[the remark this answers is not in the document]
jonaslagoni: Nice catch @smoya 👍 

> TBH is not an issue with this PR per se. The way we are exporting the schemas is not helping. I think we could introduce the following separation in next-major branch:

Fine for me 👍

smoya (approved):
LGTM! 🚀🌔

Note: Some follow up work has to be done in `next-major` branch. https://github.com/asyncapi/spec-json-schemas/pull/298#discussion_r1146137642

jonaslagoni:
This PR is not targeting a next branch nor defined as breaking. This cannot be merged in it's current state.

smoya:
> This PR is not targeting a next branch nor defined as breaking. This cannot be merged in it's current state.

Wondering if we shall consider this a breaking change.

jonaslagoni:
@smoya you just started it breaks the parser 😄

smoya:
> @smoya you just started it breaks the parser 😄

Afaik only the 2.0 version which has not been released yet.

smoya:
@jonaslagoni is your concern still there?

derberg:
@smoya @jonaslagoni 

Hey folks, this PR targets `master`. It is not breaking anything as final solution we came up with was that we actually need an alternative schemas, so this PR is not making any changes to existing schemas. This PR extends npm package with access to new alternative schemas, so it is just a `feat`

> This PR is not targeting a next branch nor defined as breaking. This cannot be merged in it's current state.

so why you think it cannot be merged?

jonaslagoni:
@derberg 

The comment is after @smoya comment about it breaking the parser.

> This change, once ported to next-major branch, will make https://github.com/asyncapi/parser-js/blob/next-major/src/schema-parser/asyncapi-schema-parser.ts#L38-L53 to provide wrong types.

The problem is we don't have a clear definition of what a breaking change means 😆 In this case it is something indirect unintended side-effect that is breaking. I.e. we can no longer iterate over what index exports to know all versions.

So yea, your call 😄

smoya:
> > The comment is after @smoya comment https://github.com/asyncapi/spec-json-schemas/pull/298#discussion_r1146137642.
> 
> sorry, I completely missed it, saw @smoya accepted so ignored the rest.
> 
> > Anyway, we must make a decision, and I see to reach a consensus, following the most conservative solution can be considered fair: to release a new major version.
> 
> well there is not much to discuss really. As you described in https://github.com/asyncapi/spec-json-schemas/pull/298#discussion_r1146137642 this PR will break parser, then it will most probably break other integrations. If we iterate over what is exported to identify all supported spec versions, then others most probably do it too.
> 
> I will update this PR so it follows https://github.com/asyncapi/spec-json-schemas/issues/367 and mark this PR as `feat!:` to release major.
> 

Cool. I can work on that PR tomorrow morning. I'll ping you once is done.

smoya:
> I will update this PR so it follows #367 and mark this PR as `feat!:` to release major.

I understood wrongly and wrote a comment but removed it. Now I get you will include the work for https://github.com/asyncapi/spec-json-schemas/issues/367 in **this** PR 👍 Awesome.

derberg:
@jonaslagoni @smoya I changed the package exports, have a look:
- prefix for major release updated
- description of PR updated

jonaslagoni (approved):
If you dont want 2 major version changes within 3 months, you could merge this into `next-major-spec`. But just a suggestion 🙂

derberg:
I do not mind having major every month if that is required 😃 we should not be afraid of major releases in libraries

smoya:
> If you dont want 2 major version changes within 3 months, you could merge this into `next-major-spec`. But just a suggestion 🙂

I agree. And keep advocating for that over long living PR's targeting `master`

smoya:
> I do not mind having major every month if that is required 😃 we should not be afraid of major releases in libraries

I'm sure releasing few many major versions in a short period of time can be taken as negative because few reasons. Completely valid for a new discussion but still worth to share IMHO:

- Can create the perception of instability, like the package is not mature enough.
- Contributors can feel disconnected from the repo when realizing few major versions happened without them to even realize.
- Annoying users to update too many times in order to keep up with big changes.

derberg:
@fmvilas I need your approval as you requested changes in the past

@dalelane @char0n have a look as we will release major here

derberg:
@fmvilas I need your approval as you requested changes in the past

@dalelane @char0n have a look as we will release major here

fmvilas (approved):

derberg:
@smoya please approve again, I had to solve merge conflict with readme cause by new breaking changes section. No more changes

derberg:
@smoya @fmvilas one approve needed

smoya (approved):

PR #312, opened 2023-02-02
feat: support multiple binding versions

jonaslagoni:
**Description**

This PR changes a few things:

1. Adds support for ONLY allowing a specific range of binding versions
2. Adds support for defaulting to the `latest` version of the bindings
3. Adds support for explicitly only check the appropriate binding schema when `bindingVersion` is explicitly defined.
4. Now only validate a binding if needed instead of all bindings, which could cause issues when reporting back errors.
5. Add missing message binding for AMQP
6. Add missing server binding for MQTT5
7. Add missing operation binding for HTTP
8. Fix the wrong version for MQTT5 server schema
9. Fix incorrect regex for pattern properties for Kafka 0.4 and MQTT5

Here is how the binding schema changes work (example with mqtt5 bindings):

```
    "mqtt5": {
      "properties": {
        "bindingVersion": {
          "enum": ["0.2.0"]
        }
      },
```
Is added to ensure only a specific range of binding versions (if the property `bindingVersion` is set) are valid i.e. it gives you an error if you use `"bindingVersion": "0.3.0",`, while still allowing no values.

```
      "allOf": [
        {
          "description": "If no bindingVersion specified, use the latest binding",
          "if": {
            "not": {
              "required": [
                "bindingVersion"
              ]
            }
          },
          "then": {
            "$ref": "http://asyncapi.com/bindings/mqtt5/0.2.0/server.json"
          }
        },
```

The first entry in the `allOf` is for matching the default version, i.e. when no specific `bindingVersion` has been sat. If so it defaults to the newest, in this case, version 0.2.0. 

```
        { 
          "if": {
            "required": [ "bindingVersion" ],
            "properties": {
              "bindingVersion": {
                "const": "0.2.0"
              }
            }
          }, 
          "then": {
            "$ref": "http://asyncapi.com/bindings/mqtt5/0.2.0/server.json"
          }
        }
```

The next entry in the `allOf` includes the validation for when the user explicitly states a binding version, i.e. if the property `bindingVersion` has been sat, and is of value `0.2.0` we validate against `http://asyncapi.com/bindings/mqtt5/0.2.0/server.json`.


jonaslagoni:
Need to double check it validates correctly.

bindings/kafka/0.4.0/channel.json
```diff
@@ -6,7 +6,7 @@
   "type": "object",
   "additionalProperties": false,
   "patternProperties": {
-    "^x-[\\w\\d\\.\\-\\_]+$": {
+    "^x-[\\w\\d\\.\\x2d_]+$": {
```
dalelane, line 9: Sorry, I'm struggling to parse this - what is the aim of this change?

jonaslagoni: If I recall correctly it was related to making the pattern property `unicode ECMA-262 compliant` cc @smoya 

It was introduced here: https://github.com/asyncapi/spec-json-schemas/pull/145

This is the regex format JSON Schema dictates: https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01#section-7.3.8

jonaslagoni: This is what all spec and binding schema files use btw 🙂

dalelane (dismissed):

char0n (approved):
LGTM, nice work :tada:

derberg (approved):

jonaslagoni:
Had to update with latest changes, just need a single review 🙂

char0n (approved):

jonaslagoni:
Thanks @char0n!

/rtm

jonaslagoni:
@derberg do you know why merging is blocked? https://github.com/asyncapi/spec-json-schemas/actions/runs/4125615904/jobs/7126408523

dalelane (approved):

jonaslagoni:
Lets try again, thanks @dalelane 😄 
/rtm

PR #329, opened 2023-02-08
feat: add host and pathname fields in server object

magicmatatjahu:
**Description**

Add `host` and `pathname` fields in server object.

**Related issue(s)**
Part of https://github.com/asyncapi/spec/pull/888

### DIFF
--- a/definitions/3.0.0/server.json
+++ b/definitions/3.0.0/server.json
@@ -2,7 +2,7 @@
   "type": "object",
   "description": "An object representing a Server.",
   "required": [
-    "url",
+    "host",
     "protocol"
   ],
   "additionalProperties": false,
@@ -12,8 +12,13 @@
     }
   },
   "properties": {
-    "url": {
-      "type": "string"
+    "host": {
+      "type": "string",
+      "description": "The server host name."
+    },
+    "pathname": {
+      "type": "string",
+      "description": "The path to a resource in the host."
     },
     "title": {
       "type": "string",


fmvilas (approved):

char0n (approved):
LGTM. Nice job!

magicmatatjahu:
@fmvilas @char0n Thanks! However I think that we should add some regex (or format) to validate the proper host and pathname, not treat fields as normal strings. WDYT? The problem is that I will (probably) learn Regex again 😆

char0n:
@magicmatatjahu we theoretically can, here are the representation in pseudo context-free grammar for host and pathname.

**host**:
```
host:
      ‘[‘ ipv6-address ‘]’
      ipv4-or-reg-name

ipv6-address:
      (any input-character except ‘/’ and ‘]’)+
ipv4-or-reg-name:
      (any input-character except ‘/’, ‘:’, ‘?’ and ‘#’)+
```

**pathname**

As far as I understand this is https://url.spec.whatwg.org/#dom-url-pathname without `search` part

```
absolute-path:
      ‘/’
      (path-separator segment)+

path-separator:
      ‘/’
      new-line-with-indent

segment:
      (any input-character except ‘/’, ‘?’ and ‘#’)*
```

It should be possible to convert these context-free grammars into regular expressions.

dalelane (approved):

fmvilas:
> @fmvilas @char0n Thanks! However I think that we should add some regex (or format) to validate the proper host and pathname, not treat fields as normal strings. WDYT? The problem is that I will (probably) learn Regex again 😆

I would create another issue for this. This is the kind of task that can quickly become a rabbithole.

magicmatatjahu:
@fmvilas Ok, I created issue https://github.com/asyncapi/spec-json-schemas/issues/336 I'm merging that PR.

PR #333, opened 2023-02-09
fix: message can have nested oneOf

### LINKED ISSUE #332: Message object allows nested `oneOf`
#### Describe the bug
It is possible to define a message such as:

```
      message:
        oneOf:
          - name: LightMeasured
            payload:
              type: string
          - oneOf: 
            - name: LightMeasured2
              payload:
                type: string
```

and having the schema valid, this should not be the case.

#### How to Reproduce
```
asyncapi: 2.6.0
info:
  title: Streetlights API
  version: 1.0.0
  description: |
    The Smartylighting Streetlights API allows you
    to remotely manage the city lights.
  license:
    name: Apache 2.0
    url: 'https://www.apache.org/licenses/LICENSE-2.0'
servers:
  mosquitto:
    url: 'mqtt://test.mosquitto.org'
    protocol: mqtt
channels:
  light/measured:
    publish:
      summary: >-
        Inform about environmental lighting conditions for a particular
        streetlight.
      operationId: onLightMeasured
      message:
        oneOf:
          - name: LightMeasured
            payload:
              type: string
          - oneOf: 
            - name: LightMeasured2
              payload:
                type: string
```

#### Expected behavior
I was expected the validation of said AsyncAPI file to be invalid instead of valid.

jonaslagoni:
**Description**
This PR fixes the nested `oneOf` problem.

Once we agree this is the fix, then I will apply it for the rest of the versions as well.

I think this will be a breaking change 🧐 

Solves #332


magicmatatjahu (changes requested):
Please also remember to update references to new `messageObject` in all places like https://github.com/asyncapi/spec-json-schemas/blob/8c9f3d72fa20d82fa7045ca5cec8e318d42e8e6e/definitions/3.0.0/channelMessages.json and also in components.

However, in 3.0.0 we don't have such a oneOf messages, we have objects like `channel.messages` or `operation.messages` (which is a array of messages) so your change should be done only for 2.x.x :) We have a "bug" in 3.0.0 schemas.

jonaslagoni:
> Please also remember to update references to new messageObject in all places like https://github.com/asyncapi/spec-json-schemas/blob/8c9f3d72fa20d82fa7045ca5cec8e318d42e8e6e/definitions/3.0.0/channelMessages.json and also in components.

Those should not be changed in this PR as it has nothing to do with this bug fix 🙂 

@magicmatatjahu lets make further changes to the schema files regarding 3.0 after this one, lets's not mix multiple different changes in one.

jonaslagoni:
But yea, components need to change 👍

jonaslagoni:
@magicmatatjahu never mind, thats actually another bug 🤣 Gonna do that in another PR after this one

jonaslagoni:
Created a separate issue for it: https://github.com/asyncapi/spec-json-schemas/issues/335

definitions/3.0.0/messageObject.json
```diff
@@ -0,0 +1,159 @@
...
+              }
+            ]
+          }
+        ]
+      }
+    }
+  },
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "http://asyncapi.com/definitions/3.0.0/messageObject.json"
+}
```
MihanixA, line 159: missing EOF

MihanixA (changes requested):

definitions/3.0.0/messageObject.json
jonaslagoni: It is just a notice from GitHub, incase it's relevant for your setup to have newlines at the end of files, in this case I am pretty sure it's not 🙂

derberg (approved):

Issue #335, opened 2023-02-09
Component messages can contain oneOf


jonaslagoni:
#### Describe the bug
According to the spec, a message in the component section should not be able to be defined with oneOf, but at the moment the following is valid:
```
components:
  messages:
    myMessage:
      oneOf:
        - name: LightMeasured
          payload:
            type: string
```

#### Expected behavior
Expected a validator to mark this as invalid input.

dalelane:
Is there more to do here?

jonaslagoni:
Nope, lets close this 😄

PR #345, opened 2023-02-22
fix: oneOf should not be allowed in message components

jonaslagoni:
**Description**
This PR fixes that it currently possible to define messages with oneOf in components when the spec clearly doesn't allow it.

Currently allowed, but should not:
```
components:
  messages:
    UserSignedUp:
      oneOf:
        - payload:
            type: object
            properties:
              displayName:
                type: string
                description: Name of the user
              email:
                type: string
                format: email
                description: Email of the user
```

**Related issue(s)**
Fixes https://github.com/asyncapi/spec-json-schemas/issues/335

### DIFF
--- a/definitions/3.0.0/components.json
+++ b/definitions/3.0.0/components.json
@@ -21,7 +21,19 @@
       "$ref": "http://asyncapi.com/definitions/3.0.0/serverVariables.json"
     },
     "messages": {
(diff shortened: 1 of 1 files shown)


dalelane (approved):

jonaslagoni:
Thanks @dalelane 👍

PR #347, opened 2023-02-22
feat: add AVRO schema validation

jonaslagoni:
**Description**
This PR adds AVRO validation for message payloads that uses the schema formats:
```
"application/vnd.apache.avro;version=1.9.0",
"application/vnd.apache.avro+json;version=1.9.0",
"application/vnd.apache.avro+yaml;version=1.9.0"
```

Based on the recommended schema formats: https://www.asyncapi.com/docs/reference/specification/v2.6.0#messageObjectSchemaFormatTable

**Related issue(s)**
Blocked by https://github.com/asyncapi/spec-json-schemas/pull/334


jonaslagoni:
This PR is ready to be reviewed as https://github.com/asyncapi/spec-json-schemas/pull/334 has been merged.

Think you have to re-run the macos test.

jonaslagoni:
See https://github.com/asyncapi/spec-json-schemas/pull/350

definitions/3.0.0/avroSchema_v1.json
```diff
@@ -0,0 +1,317 @@
+{
```
smoya, line 1: Out of curiosity, where did you get this schema from? https://json.schemastore.org/avro-avsc.json ? I couldn't see where it says it is 1.9 and not 1.10 or similar, that's why asking.

jonaslagoni: I am not 100% sure, but because the AsyncAPI example is valid, I did not look further into it 😄

smoya: So we are "maybe" giving support to Avro > 9. We recommend Avro 9 in our docs, but not sure it is a strong req.

dalelane (approved):

definitions/3.0.0/avroSchema_v1.json
smoya: @dalelane as you are more into Avro world, do you think this is important?

smoya (approved):
Even though https://github.com/asyncapi/spec-json-schemas/pull/347#discussion_r1135224631,

LGTM!

PR #350, opened 2023-03-13
test: increase timeout that periodically fails

jonaslagoni:
**Description**
This PR increases the timeout for the schema tests as we started to see spotted failed attempts in CI runs.


derberg (approved):

jonaslagoni:
Thanks @derberg ⚡

PR #363, opened 2023-03-16
feat: add Avro 1.9 Schema 

jonaslagoni:
**Description**
As done in https://github.com/asyncapi/spec-json-schemas/pull/347 this PR adds the Avro Schema object for all v2 versions.

It is quite a few duplicate schemas... But there is currently no setup to share schemas between versions... Would be great to have at some point. See https://github.com/asyncapi/spec-json-schemas/issues/364


dalelane (approved):
LGTM - as highlighted, the duplication is unfortunate but something we can hopefully improve in future

derberg (approved):

PR #370, opened 2023-03-31
feat!: allow multi format schemas

smoya:
**Description**

This PR contains all the required changes for the following new Spec v3.0.0 feature https://github.com/asyncapi/spec/issues/622. 
A [WIP PR](https://github.com/asyncapi/spec/pull/910) containing changes on the spec has been created by @GreenRover.

The changes made in this PR apply to:

- Message Payload (including examples) both in the root document and in components.
- Schemas in components

**Related issue(s)**
https://github.com/asyncapi/spec/issues/622


GreenRover:
I found a point. The `MessagePayloadObject` can not be fined below `components`. Can you add this.

jonaslagoni:
@smoya I am assuming you continue this when you get back the 19 th, otherwise let me know so I can help push this over the finish line 🙂

smoya:
> @smoya I am assuming you continue this when you get back the 19 th, otherwise let me know so I can help push this over the finish line 🙂

You are right assuming this. I'm gonna work on this.

smoya:
> I found a point. The `MessagePayloadObject` can not be fined below `components`. Can you add this.

@GreenRover Sorry for the delay, but I can't find where we discussed about having those under components. We already have `schemas` so what's the need for adding this new wrapper under components as well?

cc @derberg @jonaslagoni

definitions/3.0.0/messagePayloadObject.json
```diff
@@ -0,0 +1,147 @@
...
+          "properties": {
+            "schema": {
+              "$ref": "http://asyncapi.com/definitions/3.0.0/avroSchema_v1.json"
+            }
+          }
+        }
+      }
+    ],
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "http://asyncapi.com/definitions/3.0.0/messagePayloadObject.json"
```
jonaslagoni, line 146: I would call it something similar to `MultiFormatSchemaObject` that is unrelated to `payload`.

jonaslagoni:
> @GreenRover Sorry for the delay, but I can't find where we discussed about having those under components. We already have schemas so what's the need for adding this new wrapper under components as well?

`components.schemas` needs to change as well yea 🙂

definitions/3.0.0/messageObject.json
```diff
@@ -114,7 +113,9 @@
           "headers": {
             "type": "object"
           },
-          "payload": {}
+          "payload": {
+            "$ref": "http://asyncapi.com/definitions/3.0.0/messagePayloadObject.json"
+          }
```
jonaslagoni, line 118: This needs to be a union of the schema+format, just schema, or reference.

smoya:
@jonaslagoni I made the changes you suggested. Additionally, I removed the `object` suffix as I don't find it necessary anymore.
The multiFormatSchemas are added into the `schemas.json` now. 

BTW, I tested the compiled schema but I expected the following to give an error because the schema is not a valid avro but it doesn't. I will check the reason why.

https://www.jsonschemavalidator.net/s/aSOFbwxD

GreenRover:
> > @GreenRover Sorry for the delay, but I can't find where we discussed about having those under components. We already have schemas so what's the need for adding this new wrapper under components as well?
> 
> `components.schemas` needs to change as well yea 🙂

There is no direct need. May be reusability. 
But we never discussed this fact.

smoya:
> > > @GreenRover Sorry for the delay, but I can't find where we discussed about having those under components. We already have schemas so what's the need for adding this new wrapper under components as well?
> > 
> > 
> > `components.schemas` needs to change as well yea 🙂
> 
> There is no direct need. May be reusability. But we never discussed this fact.

Yeah makes sense. I also forget we added in that way in the spec so no need to even discuss about it 👍

smoya:
> BTW, I tested the compiled schema but I expected the following to give an error because the schema is not a valid avro but it doesn't. I will check the reason why.

I discard a problem on `avroSchema_v1.json` since [I tested it isolated](https://www.jsonschemavalidator.net/s/22EXRM4k) and seems to work. It should be in my changes I guess.

smoya:
The point is that now the following `oneOf` makes schemas to match both `schema.json` and `multiFormatSchema.json` whenever the keyword `schema` is found.

```json
 "oneOf": [
      {
      },
      {
      },
      {
        "$ref": "http://asyncapi.com/definitions/3.0.0/multiFormatSchema.json"
      }
    ]
```

I will create a wrapper on top of that and use if/then and whenever that keyword is present, will use the multiFormat or rather the schema.

smoya:
> I will create a wrapper on top of that and use if/then and whenever that keyword is present, will use the multiFormat or rather the schema.

@jonaslagoni It's done now and it's called `anySchema.json`

smoya:
Btw, I can adapt this code to fulfill with the naming used in the [Parser-JS PR](https://github.com/asyncapi/parser-js/pull/785/files#diff-794c643afbdb7b12abdb1ec0ee2a07f36407ce5bd797077baa20fa03afbdf4a2R405) @jonaslagoni created. Or we can agree on common names. I'm open to anything.

fmvilas (approved):
LGTM Although the title is confusing. Shouldn't it be `allow defining schemaFormat at schema level`?

smoya:
> LGTM Although the title is confusing. Shouldn't it be `allow defining schemaFormat at schema level`?

Yeah, the title wasn't helping. I changed it for something better, using the ubiquitous language in place.

smoya:
@GreenRover I would love to have your review, if you don't mind,  as you were the champion of this feature. 🙏

smoya:
I've added support for multi format schemas in headers. Ready to review

definitions/3.0.0/multiFormatSchema.json
```diff
@@ -0,0 +1,147 @@
...
+            }
+          }
+        }
+      },
+      {
+        "if": {
+          "required": [
+            "schemaFormat"
+          ],
+          "properties": {
```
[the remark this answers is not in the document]
smoya: If we need to add official support for protobuf, indeed this should be done inside this document by adding a new "if" section. But afaik  Otherwise, it is still supported since the schema for the object itself allows it since the `schema` has no JSON Schema definition:

```json
"properties": {
      "schemaFormat": {
        "type": "string"
      },
      "schema": {}
    },
```

Meaning the following should be valid: 

```json
"components": {
    "schemas": {
      "myProtobufSchema": {
        "schemaFormat": "application/vnd.google.protobuf;version=3",
        "schema": "message Point {\n    required int32 x = 1;\n    required int32 y = 2;\n    optional string label = 3;\n}\n\nmessage Line {\n    required Point start = 1;\n    required Point end = 2;\n    optional string label = 3;\n}\n"
      }
    }
  }
```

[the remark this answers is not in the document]
jonaslagoni: the `"schema": {}` is the same as `"schema": true`, so everything is allowed.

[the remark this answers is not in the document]
smoya: Testing, I noticed there is a bug somewhere in my current implementation where it is allowed to set a string schema even when setting the schemaFormat to be an AsyncAPI schema. 

The following example demonstrates it; where the `testinline` schema appears as invalid (as I expect) but the `test` schema is valid (unexpected). 

```json
{
  "asyncapi": "3.0.0",
  "info": {
    "title": "Example",
    "version": "0.1.0"
  },
  "components": {
    "schemas": {
      "test": {
        "schemaFormat": "application/vnd.aai.asyncapi+json;version=3.0.0",
        "schema": "message Point {\n    required int32 x = 1;\n    required int32 y = 2;\n    optional string label = 3;\n}\n\nmessage Line {\n    required Point start = 1;\n    required Point end = 2;\n    optional string label = 3;\n}\n"
      },
      "testInline": "message Point {\n    required int32 x = 1;\n    required int32 y = 2;\n    optional string label = 3;\n}\n\nmessage Line {\n    required Point start = 1;\n    required Point end = 2;\n    optional string label = 3;\n}\n"
    }
  }
}
```

[Open in JSON Schema validator](https://www.jsonschemavalidator.net/s/8bpDkYR3)

I don't get why since both are validating against `schema.json`. I will need to check what's the issue about.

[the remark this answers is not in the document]
jonaslagoni: @smoya remember to change the components to be `anySchema.json` not `schema.json` 😄

definitions/3.0.0/schemas.json
```diff
@@ -1,9 +1,9 @@
 {
+  "description": "An object representing multiple schemas.",
```
jonaslagoni, line 2: This file is actually no longer used.

jonaslagoni (changes requested):

definitions/3.0.0/schemas.json
smoya: good catch

definitions/3.0.0/multiFormatSchema.json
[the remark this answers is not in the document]
smoya: oh-no-facepalm-gif

definitions/3.0.0/components.json
```diff
@@ -12,14 +12,7 @@
...
         "^[\\w\\d\\.\\-_]+$": {
-          "oneOf": [
-            {
-              "$ref": "http://asyncapi.com/definitions/3.0.0/Reference.json"
-            },
-            {
-              "$ref": "http://asyncapi.com/definitions/3.0.0/schema.json"
-            }
-          ]
+          "$ref": "http://asyncapi.com/definitions/3.0.0/anySchema.json"
```
jonaslagoni, line 15: Dont remove reference, both has to be present, otherwise a schema component cannot be a reference.

jonaslagoni (changes requested):
One comment

smoya:
@jonaslagoni I had to add [this change](https://github.com/asyncapi/spec-json-schemas/pull/370/commits/266fd07f2c5db3446554c461bde290924fae1413) in order to allow values such as bool to be valid schemas when matched with `multiFormatSchema.json`. Otherwise, the following will be invalid:

```json
"components": {
    "schemas": {
      "test": {
        "schemaFormat": "application/vnd.google.protobuf;version=3",
        "schema": "message Point {\n    required int32 x = 1;\n    required int32 y = 2;\n    optional string label = 3;\n}\n\nmessage Line {\n    required Point start = 1;\n    required Point end = 2;\n    optional string label = 3;\n}\n"
      },
      "test2": true,
      "test3": {
        "type": "object",
        "properties": {
          "foo": true
        }
      },
      "test4": {
        "schemaFormat": "application/vnd.aai.asyncapi;version=3.0.0",
        "schema": true
      }
    }
  }
```

Do you believe there is a better/smarter way of doing that?

definitions/3.0.0/anySchema.json
```diff
@@ -0,0 +1,21 @@
+{
+  "allOf": [
```
fmvilas, line 2: Is `allOf` needed here? 🤔

fmvilas (changes requested):

fmvilas (approved):
🚢 🇮🇹

PR #389, opened 2023-04-27
feat!: release version 5.0.0

derberg:
https://github.com/asyncapi/spec-json-schemas/pull/386 did work

I finally learned why. There release config was in a wrong location so `semantic-release` applied defaults, which is for conventional commits it used `angular` way that do not respect `feat!` or `fix!` but requires explicit `BREAKING CHANGE:`

now this release config is globally fixed everywhere and in the root of the repo


smoya (approved):
I like your perseverance.

PR #406, opened 2023-06-30
Merge `master` into `next-major-spec` + fix bug in bundler

smoya:
**Description**

This PR, besides merging `master` into `next-major-spec`, fixes a bug in the bundler with `$ref` pointing to definitions without `.json` extension. The bugfix adds a new use case when parsing the bundled schema and finding `$ref` pointing to things like `"$ref": "http://asyncapi.com/definitions/3.0.0/operation.json#/properties/title"`. `OperationTrait` has some of those now in `3.0.0`: see https://github.com/asyncapi/spec-json-schemas/blob/next-major-spec/definitions/3.0.0/operationTrait.json#L11.

> **Warning**
> Manually merge and not rebase!


smoya:
@derberg @jonaslagoni  I need some help debugging what's the issue with the produced `3.0.0-without-$id.json`. 
Here you can find the validator error: https://www.jsonschemavalidator.net/s/3OO3lvSb

> Could not resolve schema reference '#/operation#/properties/title'. Path 'definitions.operationTrait.properties.title', line 4675, position 26.

derberg:
Interesting, it is only with `operationTrait`, not a serious stuff imho

the oryginal schema has `http://asyncapi.com/definitions/3.0.0/operation.json#/properties/title` and the no-id bundler is dummy and replace `http://asyncapi.com/definitions/3.0.0/operation.json` with a local reference. So simple fix is just have an if that if there is a url with anchors, then not only `http://asyncapi.com/definitions/3.0.0/operation.json` but `http://asyncapi.com/definitions/3.0.0/operation.json#` should be replaced new local reference so it looks like `"$ref": "#/definitions/operation/properties/title"`

smoya:
> "$ref": "#/definitions/operation/properties/title"

@derberg The issue is with JSON pointer here and the fact we should scape `/` characters with `~1`, which doesn't look pretty good from DX perspective.

I decided to use `-` instead now (See the last commit I pushed). For example,`http://asyncapi.com/bindings/http/0.1.0/message.json` in regular file is `bindings-http-0.1.0-message` in `-without-$id` file.

I also added support for testing and validating the schemas without $id which was not added before.

derberg:
@smoya I don't get why you need `~1`? we use it in parser-js only if we need to handle encoding of `/` in the channel names that may have it. But in case of stuff like `"#/definitions/operation/properties/title"` there is no need to do any tilde encoding really 🤔

smoya:
> @smoya I don't get why you need `~1`? we use it in parser-js only if we need to handle encoding of `/` in the channel names that may have it. But in case of stuff like `"#/definitions/operation/properties/title"` there is no need to do any tilde encoding really 🤔

My bad for the confusing explanation. Gonna try again.


### 1. Bug with `#/operation#/properties/title`
The issue was that the extra `#` was not required. That's fixed, and now it is `#/operation/properties/title`.

### 2. Bug with definition ids with `/` characters
Ids can have extra `/` characters. That's the case for the bindings, for example, where you can find things like `http://asyncapi.com/bindings/mqtt/0.1.0/operation.json`. 
If we convert that to `bindings/mqtt/0.1.0/operation`, then references will need to use `~`, such as `"#/definitions/bindings~1mqtt~10.1.0~1operation"`. Otherwise, the JSON Pointer will try to find in the path `definitions.bindings.mqtt.0.1.0.operation`. Meaning it will only work if a definition like the following is in the file (which is not our case):

```json
 "bindings": {
            "mqtt": {
                "0.1.0": {
                    "operation": {
                        "type": "object"
                    }
                }   
            }
        },
```

derberg:
> "0.1.0

ah yeah, makes a lot of sense

derberg (approved):

PR #421, opened 2023-08-07
fix: prepare `next-major-spec` for release

jonaslagoni:
**Description**
This PR adds the final changes to `next-major-spec` that is needed before release, this includes:

- Added migration guide for v6
- Added migration guide for v5 (was missing)
- Added v3 schemas to the all-schema store

**Related issue(s)**
Blocking #208


schemas/all.schema-store.json
```diff
@@ -172,6 +172,20 @@
...
+        {
+          "properties": {
+            "asyncapi": {
+              "const": "3.0.0"
+            }
+          }
+        },
+        {
+          "$ref": "http://asyncapi.com/schema-store/3.0.0-without-$id.json"
+        }
```
jonaslagoni, line 187: Not sure whether we need to do this now?

jonaslagoni:
Adapted this PR to match the desire to not release before the spec.

migrations/migrate-to-version-6.md
```diff
@@ -0,0 +1,3 @@
+# Migrating to version 6
+
+In version 6 and onwards, all bindings will now be validated along side the specification. This means that documents that was valid before might not be valid anymore if the bindings in their document is incorrect.
```
derberg, line 3: regarding the statement here, bindings are only integrated starting from AsyncAPI 3.0, not with older schemas?

jonaslagoni: You are right 🤔 

Hmm, why did we release it as v6 then? 😆 I cant find any other breaking changes 🤔

derberg: you need to introduce breaking changes as agreed here -> https://github.com/asyncapi/spec-json-schemas/pull/208#issuecomment-1667567530

jonaslagoni: Just so I read it correctly, we expose it as this:
>module.exports = {
    'schemas': {
      ...
    },
    'schemasWithoutId': {
      ...
    },
    'preReleases': {
      ...
     },
    'preReleasesWithoutId': {
      ...
     }
};

Right? 🤔

jonaslagoni: Or did I misunderstand your comment @derberg 🤔

derberg: I see that in linked comment I provided YT link to a bad time.
Fixed: https://youtu.be/WW5sLVHa0b8?t=1738

So we do it the way you advocated for, so no `preReleases` keys, all goes to `schemas`, but first time we do it as major:

- so we need in readme clear info that we will add future releases directly to package as minors, that until spec is released, they will be vulnerable for breaking changes (schemas). Best a separate and clear section
- migration guide could have this info - the one from this PR
- would be good if the main PR that we merge, also has this info in description and even PR title, so you know, peopole that go to GitHub Release notes know what to expect


you also described it in: https://youtu.be/WW5sLVHa0b8?t=1937

smoya: > * so we need in readme clear info that we will add future releases directly to package as minors, that until spec is released, they will be vulnerable for breaking changes (schemas). Best a separate and clear section

I already mentioned this somewhere else I can't recall, but what about tools that rely on that list as a list of stable spec versions? The closest example is the Parser-JS, which uses the latest key as the latest spec version, and have several rules based on that (like a warning suggesting you are not using the latest version of the spec, etc) https://github.com/asyncapi/parser-js/blob/next-major-spec/src/constants.ts#L22

jonaslagoni: They have to whitelist if they do not want each new version

smoya (changes requested):
@derberg @jonaslagoni Would you mind answering (maybe it has been answered already somewhere else) the following https://github.com/asyncapi/spec-json-schemas/pull/421/files/b43d3d650d24ab99c9e2a7436d1ba7d1adb60bc9..340e842a3a27498361022e743b4239fbb77d593e#r1350501180 before moving forward?

Thank you 🙏

migrations/migrate-to-version-6.md
smoya: I don't think it is a solution to say "just whitelist". We are assuming people assumptions 🤔. I don't really understand the (strong) reason behind all of that.

jonaslagoni: We are not assuming anything 😄 we are declaring an approach on how the schema list works that people must follow.

If it does not work out, we can always change the behaviour again. But I dont think its gonna be a problem.

smoya: > We are not assuming anything 

We do. We assume tools won't have an undesired behavior when adding a schema that is under development and open for breaking changes. 
Even further, we assume people must know that that particular new major version is not released yet.

Imagine a user, generating it's first AsyncAPI document via VS Code, or any other IDE that assume the list of specs given by this package are the ones the editor can support. Imagine those users generating AsyncAPI documents with the latest (because of course, those IDE could suggest the user to use the latest version) and being invalid 5 mins after because a new prerelease is out. 

This is a simple example, but complex scenarios could happen in SaaS tools, internal projects, etc.

derberg: I don't understand, are you maybe mixing in parser use cases here?

here in spec json schema the only assumption we have is:
- you read docs before you use the library
- you read release notes before you upgrade to latest major

if in parser you decide that v3 or v4 should not be available in the parser before the release - good, then lock your code to specific spec version, like we do with package-lock for libraries, and expose other version behind feature flag if you want. But Parser has nothing to do here

migrations/migrate-to-version-6.md
```diff
@@ -0,0 +1,21 @@
...
+        '3.0.0': require('./schemas/3.0.0.json'),
+    },
+    'schemasWithoutId': {
+        ...
+        '3.0.0': require('./schemas/3.0.0-without-$id.json'),
+    }
+};
+```
+
+However, while the it's still a pre-release, the underlying schemas CAN contain breaking changes from version to version, up until the AsyncAPI specification is released. This means that one AsyncAPI document using v3 in the pre-release stage might be valid in `6.0.0`, but invalid in the `6.1.0`. This ONLY applies to pre-release schemas, and NOT regular ones that is set in stone.
```
dalelane, line 19:
```suggestion
However, while it's still a pre-release, the underlying schemas CAN contain breaking changes from version to version, up until the AsyncAPI specification is released. This means that one AsyncAPI document using v3 in the pre-release stage might be valid in `6.0.0`, but invalid in the `6.1.0`. This ONLY applies to pre-release schemas, and NOT regular ones that is set in stone.
```

fmvilas (approved):
LGTM 👍

char0n (approved):
LGTM

migrations/migrate-to-version-6.md
derberg: > Consider VSCode a user of our schemas through the schema-store.org related plugins.

I cannot imagine anyone using AsyncAPI next release for production, by editing intentionally stable 3.0 and switch to 3.1 that is not yet released, without checking docs and figuring what is the difference.

We can agree that descriptions like https://github.com/asyncapi/spec-json-schemas/blob/master/schemas/2.6.0-without-%24id.json#L3179 will clearly say that something is release candidate

You mean like `stableSchemas` in https://github.com/asyncapi/spec-json-schemas/blob/master/index.js, yeah why not

derberg:
looks like everybody agree that https://github.com/asyncapi/spec-json-schemas/pull/421#discussion_r1408275309 is a direction so @jonaslagoni you can update this PR and we can merge it before release

migrations/migrate-to-version-6.md
jonaslagoni: `stableSchemas` is the same as adding `preReleases` 😄

smoya (approved):
LGTM! 🚀🌔

migrations/migrate-to-version-6.md
smoya: Let's end up this discussion (at least on my side). I don't fully agree but I see both sides of the discussions are valid points, so let's move up with the current approach you already created and, if it doesn't work, we come back to this discussion again 👍 

Thanks for all this back and forth, totally needed in my opinion.

derberg:
oh wait, so what about below @jonaslagoni 

> looks like everybody agree that https://github.com/asyncapi/spec-json-schemas/pull/421#discussion_r1408275309 is a direction so @jonaslagoni you can update this PR and we can merge it before release

migrations/migrate-to-version-6.md
derberg: > stableSchemas is the same as adding preReleases 😄 You have to do work on each release to adapt it, so in the end if we choose one of those options we gonna have to do work regardless, which this tries to avoid.

it is not the same

`preReleases` idea was to have separate objects:
- one for official releases
- one for candidates

`stabeSchemas` next to `schemas` is:
- one object with all schemas (official and candidates)
- one object with only official releases

I think difference is big as it enables tools developers to stay stable old way, and others to not. It also enables you in parser and other tools to implement feature flags, that you will be able to use to decide if you want to pick only official or candidates too. It is much more handy, as with feature flag you are not in rush, and you can turn it off on release day or in few weeks/months after.

👆🏼 anyway `stabeSchemas` can be added later

questions is @jonaslagoni do you want to followup with what we talked about - update of readme with info that from now on `master` will contain future candidates too?

derberg: luckily I just noticed @jonaslagoni in his PR had migration guide ready accordingly so we were actually only missing the readme. This PR https://github.com/asyncapi/spec-json-schemas/pull/458 should fix it


also created a followup for `stableSchemas` idea -> https://github.com/asyncapi/spec-json-schemas/issues/459

jonaslagoni: Seems like you all worked it out 😄 👍

jonaslagoni:
All good 👍

PR #424, opened 2023-09-15
chore: exclude tools/bundler/node_modules from final package

Baldinof:
**Description**

On a fresh install of the package `node_modules` in the `tools/bundler` are included while it's only useful when publishing the package.

This will allow to save some space / bandwith when installing this package.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -3,3 +3,4 @@ test
 go.mod
 go.sum
 *.go
+tools/bundler/node_modules
--- a/.npmignore
+++ b/.npmignore
@@ -3,4 +3,4 @@ test
 go.mod
 go.sum
 *.go
-tools/bundler/node_modules
+tools/bundler


smoya:
Hi @Baldinof! Thanks for submitting this PR!

`tools/bundler` is an isolated library, that has it's own `package.json` and could be used outside this project as well.
What do you think about moving that line of gitignore to a new `.gitignore` file located under `tools/bundler` dir?

Does it make sense?

Baldinof:
Thanks for you answer @smoya :)

What do you mean by _could be used outside this project as well_? It can be installed separately?

This PR is to remove the embeddings of the node_modules in `tools/bundler` from the `@asyncapi/specs` package tarball.

I don't think adding a `.gitignore` under `tools/bundler`.

But maybe `tools/bundler` is part of the public API of this package but it does not seem to be the case (it's only used to generate files for the release).

smoya:
> Thanks for you answer @smoya :)
> 
> What do you mean by _could be used outside this project as well_? It can be installed separately?
> 
> This PR is to remove the embeddings of the node_modules in `tools/bundler` from the `@asyncapi/specs` package tarball.
> 
> I don't think adding a `.gitignore` under `tools/bundler`.
> 
> But maybe `tools/bundler` is part of the public API of this package but it does not seem to be the case (it's only used to generate files for the release).

As you mentioned, the tools/bundler is expected to be detached from the @asyncapi/specs. It is also isolated in terms of code (all code is located **only** in `tools/bundler`), that's why I suggest it needs an independent `.gitignore` file under `/tools/bundler` and not to mix with the main `.gitignore` file in the root.

Baldinof:
The code change is about `.npmignore`, not `.gitignore` to ignore files in the npm package tarball.

Node modules are already ignored for git thanks to [this](https://github.com/asyncapi/spec-json-schemas/blob/master/.gitignore#L1)

smoya:
> The code change is about `.npmignore`, not `.gitignore` to ignore files in the npm package tarball.
> 
> Node modules are already ignored for git thanks to [this](https://github.com/asyncapi/spec-json-schemas/blob/master/.gitignore#L1)

omg 🤦 I swear I read .gitignore in the changeset... shame on me! Excuse me for making you waste your time 🙏

smoya:
> The code change is about `.npmignore`, not `.gitignore` to ignore files in the npm package tarball.
> 
> Node modules are already ignored for git thanks to [this](https://github.com/asyncapi/spec-json-schemas/blob/master/.gitignore#L1)

omg 🤦 I swear I read .gitignore in the changeset... shame on me! Excuse me for making you waste your time on that matter 🙏

.npmignore
```diff
@@ -3,3 +3,4 @@ test
 go.mod
 go.sum
 *.go
+tools/bundler/node_modules
```
smoya, line 6:
```suggestion
tools/bundler
```

Bundler is only use as npm script

smoya (changes requested):

.npmignore
Baldinof: Even better :)

Baldinof:
> omg 🤦 I swear I read .gitignore in the changeset... shame on me! Excuse me for making you waste your time on that matter 🙏

No worries :)

smoya (approved):
LGTM!🙌

jonaslagoni (approved):
Thanks @Baldinof 👌

PR #429, opened 2023-10-05
fix: remove trailing comma from kafka and googlepubsub bindings for messages

AceTheCreator:
Remove trailing commas from some bindings.

Bundling was broken because of it. CI before did not indicate any issues because of Node version. Now package lock is at version 2 and we will use Node 16 instead of 14 - no more issue with CI, in future we will see errors

### DIFF
--- a/bindings/googlepubsub/0.2.0/message.json
+++ b/bindings/googlepubsub/0.2.0/message.json
@@ -38,12 +38,12 @@
   "examples": [
     {
       "schema": {
-        "name": "projects/your-project-id/schemas/your-avro-schema-id"
+        "name": "projects/your-project-id/schemas/your-avro-schema-id",
       }
     },
     {
       "schema": {
-        "name": "projects/your-project-id/schemas/your-protobuf-schema-id"
+        "name": "projects/your-project-id/schemas/your-protobuf-schema-id",
       }
     }
   ]
--- a/bindings/googlepubsub/0.2.0/message.json
+++ b/bindings/googlepubsub/0.2.0/message.json
@@ -38,12 +38,12 @@
   "examples": [
     {
       "schema": {
-        "name": "projects/your-project-id/schemas/your-avro-schema-id",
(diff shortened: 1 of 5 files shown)


derberg (approved):
well spotted 🚀 

please make sure to bump next major for parser with latest spec package

derberg:
@dalelane hey, we need your approval cause of kafka

@jonaslagoni I see you modified codeowners -> https://github.com/asyncapi/spec-json-schemas/blob/next-major-spec/CODEOWNERS to match bindings repo, but did you discuss it with particular code owners that you added to the list? I don't think there were added to the repo. Also don't think they need to be here.

derberg:
imho we should have the same codeowners setup as in master, and if we wanna replicate setup from bindings, this needs a discussion and opt in from codeowners, as even in bindings repo, json schemas were owned by Khuda, not others

jonaslagoni:
> @dalelane hey, we need your approval cause of kafka
> 
> @jonaslagoni I see you modified codeowners -> https://github.com/asyncapi/spec-json-schemas/blob/next-major-spec/CODEOWNERS to match bindings repo, but did you discuss it with particular code owners that you added to the list? I don't think there were added to the repo. Also don't think they need to be here.

I mean we discussed it here: https://github.com/asyncapi/spec-json-schemas/pull/239#discussion_r1066822528

Think I forgot to ping codeowners afterwards yea. 

But technically, I see it the other way around, this setup just mimics what we had in bindings, so technically no new changes need approval, except the have two repos now instead of one. 

Not adding them as codeowners, would indeed need their approval IMO.

> Also don't think they need to be here.

I complete disagree, otherwise they are not really codeowners of bindings as they can only partial accept changes 🤷

smoya (approved):
LGTM! 🚀🌔

dalelane (approved):

derberg:
@jonaslagoni ah, you're completely right, sorry, forgot about that discussion.

please ping affected maintainers in https://github.com/asyncapi/bindings/issues/113 and I will invite them to this repo

PR #434, opened 2023-10-10
fix: examples of messages should not follow json schema

derberg:
https://github.com/asyncapi/spec-json-schemas/pull/370 introduced bug in Message Object JSON schema

`examples` `headers` and `payload` should not follow json schema spec

just like in [previous versions](https://github.com/asyncapi/spec-json-schemas/blob/next-major-spec/definitions/2.6.0/message.json#L118-L124) these fields need to be flexible as `payload` or `headers` are no necessarily json, could be avro for example

noticed with https://github.com/asyncapi/spec/blob/next-major-spec/examples/websocket-gemini.yml
Screenshot 2023-10-10 at 17 52 12

I checked the change locally and works like a charm

### DIFF
--- a/definitions/3.0.0/messageObject.json
+++ b/definitions/3.0.0/messageObject.json
@@ -100,10 +100,11 @@
             "description": "A brief summary of the message example."
           },
           "headers": {
-            "$ref": "http://asyncapi.com/definitions/3.0.0/anySchema.json"
+            "type": "object",
+            "description": "Schema definition of the application headers."
           },
           "payload": {
-            "$ref": "http://asyncapi.com/definitions/3.0.0/anySchema.json"
+            "description": "Definition of the message payload. It can be of any type"
           }
         }
       }
--- a/definitions/3.0.0/messageObject.json
+++ b/definitions/3.0.0/messageObject.json
@@ -101,10 +101,10 @@
           },
           "headers": {
             "type": "object",
-            "description": "Schema definition of the application headers."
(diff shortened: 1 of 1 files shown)


smoya:
There is something I should be missing because when I read the following in the spec: 

Google Chrome_TZUvXJc2@2x

- For headers field in message example object:
  > The value of this field MUST validate against the [Message Object's headers](https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.14#messageObjectHeaders) field.
- For payload field in message example object:
  > The value of this field MUST validate against the [Message Object's payload](https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.14#messageObjectPayload) field.

With that statement, I understand that it should stay as it is, with the `anySchema`, allowing then any schema (JSON Schema, Avro, etc).
Not sure what I'm missing!

derberg:
but the spec says map of string and any, so no, it should not stay as it is. As not we expect that in example of the payload, people should provide schema, while it should be an example of payload that is valid with with the schema provided in payload

now, if my payload is 
```
type: string
```

my example must be
```
type: string
```

which makes no sense as I want my example to be

```
'my example'
```

jonaslagoni (approved):
You are right, examples should not have those references 😄

smoya (approved):
LGTM! 🚀🌔 Thanks for this catch and thanks for clarifying something might sound obvious 🙌

PR #439, opened 2023-11-03
feat!: remove messageId from Message and MessageTrait objects

fmvilas:
**Description**

Removes the `messageId` property from the Message and MessageTrait objects.

**Related issue(s)**
https://github.com/asyncapi/spec/issues/978


smoya (approved):
LGTM! 🚀🌔

Issue #280, opened 2022-10-27
Extensions Catalog


Sihamtahi:
#### Reason/Context
As part of the "MVP integration of extensions catalog with AsyncAPI tools to make extension catalog useful" mentorhsip program, detailed here  https://github.com/asyncapi/extensions-catalog/issues/78.  We would like to add in this repo a part that concerns the description of extensions. In the same way as the bindings and spec. 

#### Description
In brief, An extension enables the description of functionality or protocol related features. Everyone can create their own extensions to give users more freedom on the spec. 

We need to have a JSON Schema  validation. In fact, to enable validation in the parser, we need [JSON Schema for AsyncAPI](https://github.com/asyncapi/spec-json-schemas) to contain JSON Schemas of individual extensions. 

The final objective is to have a catalog of extensions that users can browse and use. Then, as time goes by, we enhance the catalog according to the demands / needs.

derberg:
it is basically about doing what @jonaslagoni does here https://github.com/asyncapi/spec-json-schemas/pull/239
just instead of `binding` folder there will be `extensions` folder with schemas, and these will be referenced in required locations inside the main AsyncAPI JSON Schema

PR #443, opened 2023-11-08
feat: add extensions from extensions catalog to schema

sambhavgupta0705:
This PR is part of asyncapi mentorship program 
related issue: https://github.com/asyncapi/spec-json-schemas/issues/280


smoya:
@sambhavgupta0705 can you please link the issue related with this PR so people can know about it? Thanks!

sambhavgupta0705:
@smoya sorry for the late reply 
here is the issue link https://github.com/asyncapi/spec-json-schemas/issues/280

derberg (commented):
I opened a PR to your fork with enhancements. It was more complex than I thought and in depts knowledge of JSON Schema was needed.

Before you merge my PR into your fork (that will automatically update this PR):
- please familiarize with changes in bundler script. They were necessary if we are about to treat extensions the same as bindings
- info schema is not an `allOf` because I believe best practice is already defined by bindings, where bindings have separate definition schema that accommodates all bindings for a given keyword. So this is why I suggest `infoExtensions.json` that will be responsible to keep a list of all `info` extensions, so we do not "pollute" the main `info` schema. 



so yeah, proposal is that whenever there is a new extension in the catalog, for any part of the AsyncAPI document, like `server` or `operations` or whatever - extension should never be added inside the `operations` but `operations` should be modified into allOf and there needs to be a new `operationsExtensions.json` created and new schema of new extension must be listed there.

Your job now is to now add to README a section where you explain a requirement on how extension schema should be added to the repo:
- where to put extensions:
   - what folder
   - what versioning means
   - extension name in folder structure
   - schema always in file called `schema.json`
- that there needs to be a corresponding `definition` that relates to AsyncAPI object, like `info` and that it must have `Extensions` suffix. Basically point in readme to `x-x` as example
-

derberg:
@sambhavgupta0705 one more thing, changes that I propose in my PR, I tested them locally after bundling schema, converting test yaml file from repo to v3 and adding below:

```
# yaml-language-server: $schema=/Users/wookiee/sources/asyncapi-node/schemas/3.0.0-without-$id.json
asyncapi: 3.0.0
info:
  # as you can see above locally I was using bundled local schema as yaml language, and below usage of x-x was shown as valid, and when I was using numbers I was getting errors
  x-x: MyTwitterProfile
  title: Streetlights Kafka API
  version: 1.0.0
  description: |-
    The Smartylighting Streetlights API allows you to remotely manage the city
    lights.
    ### Check out its awesome features:
  
    * Turn a specific streetlight on/off 🌃  
    * Dim a specific streetlight 😎
    * Receive real-time information about environmental lighting conditions 📈
  license:
    name: Apache 2.0
    url: 'https://www.apache.org/licenses/LICENSE-2.0'
```

sambhavgupta0705:
> @sambhavgupta0705 one more thing, changes that I propose in my PR, I tested them locally after bundling schema, converting test yaml file from repo to v3 and adding below:

can you please explain this a little more whenever you get time @derberg

derberg:
@sambhavgupta0705 I mean that this is how I tested to make sure my changes produce a schema that is a valid schema

please also extend this PR by adding https://github.com/asyncapi/cli/blob/master/.sonarcloud.properties file, and making sure that whatever is in `tools` is ignored

derberg:
@sambhavgupta0705 ready for final review?

sambhavgupta0705:
@derberg yes it is

README.md
```diff
@@ -210,7 +212,17 @@ Whenever you make changes in AsyncAPI JSON Schema, you should always manually ve
    ```yaml
    # yaml-language-server: $schema=YOUR-PROJECTS-DIRECTORY/spec-json-schemas/schemas/2.6.0-without-$id.json
    ```
+   
+## Extensions
+The Extension Catalogue schema in AsyncAPI defines a standardized way to include additional metadata or custom attributes within an AsyncAPI document.
```
derberg, line 217:
```suggestion
Extensions are a way to [extend AsyncAPI specification](https://www.asyncapi.com/docs/concepts/asyncapi-document/extending-specification) with fields that are not yet defined inside the specification. To add JSON schema of the extension in this repository, you need to first make sure it is added to the [extension-catalog](https://github.com/asyncapi/extensions-catalog) repository.
```

README.md
```diff
@@ -210,7 +212,17 @@ Whenever you make changes in AsyncAPI JSON Schema, you should always manually ve
...
+### How to add extensions
+There is defined structure of adding the exrensions which is defined as followed:
```
derberg, line 219:
```suggestion
### How to add schema of the extension

```

no need to add more text

sambhavgupta0705:
Done @derberg 🚀

derberg:
@dalelane @char0n @fmvilas @smoya 

please have a look at this one - it introduces a way of adding schemas of extensions into official json schema

of course not just any random extension can be added here, but only extensions that are contributed to https://github.com/asyncapi/extensions-catalog. This should encourage community to share extensions.

and yeah, unfortunate is that the first extension we have is from old times when Twitter was Twitter, so the new extension is called `x` which looks strange and people will use it as `x-x` 😆 blame Musk

PR #445, opened 2023-11-09
fix: add operations to the components section

### LINKED ISSUE #432: operations is not allowed in components section
#### Describe the bug
Based on https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.12#componentsObject `components` section should allow `operations` in the components section.

#### How to Reproduce
parse an AsyncAPI file with:
```yaml
...
components:
  operations: {}
...
```
#### Expected behaviour
It should allow me to have operations in the components section.
I am not sure if it's a spec issue or a parser issue. 🤔

KhudaDad414:
fixes #432

### DIFF
--- a/definitions/3.0.0/components.json
+++ b/definitions/3.0.0/components.json
@@ -68,6 +68,21 @@
         }
       }
     },
+    "operations": {
+      "type": "object",
+      "patternProperties": {
+        "^[\\w\\d\\.\\-_]+$": {
+          "oneOf": [
+            {
+              "$ref": "http://asyncapi.com/definitions/3.0.0/Reference.json"
+            },
+            {
+              "$ref": "http://asyncapi.com/definitions/3.0.0/operation.json"
+            }
+          ]
+        }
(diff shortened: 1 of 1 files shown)


definitions/3.0.0/components.json
```diff
@@ -68,6 +68,21 @@
...
+    "operations": {
+      "type": "object",
+      "patternProperties": {
+        "^[\\w\\d\\.\\-_]+$": {
+          "oneOf": [
+            {
+              "$ref": "http://asyncapi.com/definitions/3.0.0/Reference.json"
+            },
+            {
+              "$ref": "http://asyncapi.com/definitions/3.0.0/operation.json"
```
char0n, line 80: I was a bit confused here because the spec references `Operation Item Object`. 

image

But it looks like it's an editorial error which I've addressed in: https://github.com/asyncapi/spec/pull/988

char0n (approved):
LGTM

PR #452, opened 2023-11-21
feat: new script and ci for JSON Schema validation

### LINKED ISSUE #436: [CI] Lint and validate JSON Schema docs
#### Reason/Context
In order to follow an standard, but also ensuring any change won't break the JSON Schema documents, I suggest we add a new lint + validate action in our GH CI workflows that run on each PR and block the PR if JSON Schema files are not valid or following our requirements.

#### Description
Consider adding a new script to the package.json scripts section, that lints and validate JSON Schema files. Something like `npm run lint` or similar. Also could be included in `npm test` call. In that way, developers could run that command before pushing code but also during the execution of CI tests.

Run the following CI tasks on the JSON Schema files on each PR. It doesn't mean all of the following should run in different actions since most of them could be done just with one tool like Spectral.

- [ ] JSON linting
- [ ] Validate JSON Schema files structure
- ~[ ] Add some rules (Spectral?) for validating things like checking all definitions come with their `examples` and `description` fields. Any others are welcome as well.~ Not included for the bounty program

#### Detailed context

- This repository holds JSON Schema documents, which are used for validating AsyncAPI documents (transformed to JSON). The parsers mainly use those JSON Schemas (`parser-js`, for instance) to validate AsyncAPI documents.
- The JSON Schemas used by the parsers and any other tool that wants to validate AsyncAPI documents are located in the `/schemas` dir. 
- Those schemas result from bundling several smaller schemas, the ones located under `/definitions.` Each of those small schemas usually refers to the validation related to just one object from the AsyncAPI spec, I.e., Message Object, Channel Object, etc.
- When people need to change those schemas, they **do not** modify them directly under `/schemas` but rather modify a particular (sometimes several of them) file under `/definitions`. That's done this way to help developers focus on what they want to change and not deal with a big file containing the whole JSON Schema doc for the entire specification.

##### Technical details that concern this issue:

- CI scripts that run on every Pull Request are [these](https://github.com/asyncapi/spec-json-schemas/blob/master/.github/workflows/if-nodejs-pr-testing.yml).
- The step that says `Run linter` is not about lining JSON Schemas but linting Javascript code. So, it has nothing to do with the work we want to do here.
- The latest step it runs is bundling all smaller schemas found under `/definitions` into the bundled big ones located in `/schemas`.
- We should validate those (`/schemas`), meaning our JSON Schema linting script should run after the generation of assets (the actual last step).

To illustrate the details above, here is a simple flow chart. Note that the script we should create is represented in the `Validate schemas from /schemas` step.

```mermaid
flowchart TD
    PR[PR] -->|Update schemas under /definitions| CI(CI / GH Actions)
    CI --> bundle(Bundle definitions into /schemas)
    bundle --> compile(Validate schemas from /schemas)
    subgraph The new action we should build
        compile -->result{Are schemas valid?}
        result -->|Yes| ok[OK - exit code 0]
        result -->|NO| ko[KO - print validation errors + exit code 1]
    end
```

AnimeshKumar923:
**Description**

- This PR introduces a new workflow that checks and validates any incoming changes in the schemas present inside the `schemas` directory.
- It also introduces a JavaScript linter.

**Tasks**
- [x] Create a JSON schema validation script. ([this is the script](https://github.com/asyncapi/spec-json-schemas/pull/452/files#diff-1ee84b0b752624e8218489fead8e59e7dd93c1841a0b80418ecfa6cde1d57497))
- [x] ~Integrate JavaScript linter. (ESLint?) Change the existing `if-nodejs-pr-testing.yml` to lint JavaScript files.~ (Will be converted into a new issue in the future)
- [x] ~Change the existing `if-nodejs-pr-testing.yml` to validate incoming schemas. (implemented [here](https://github.com/asyncapi/spec-json-schemas/pull/452/files#diff-fd411b4d11d3f5d1d4406fc43b9696941eaec0a37f53b34e04cd1adac4b8e193))~ Instead, do changes in `test` field of the script and include execution of the script from there itself (implemented [here](https://github.com/asyncapi/spec-json-schemas/commit/aeada06a8916c56a9104be48254f79ab9e1321e6))

**Related issue(s)**
Fixes #436


AnimeshKumar923:
@smoya 

- I've made the script a bit more exclusive for the `draft-04` files.
- Updated the workflow to include `lint` and `lint-schemas` steps.
- Update `package.json` dependencies and `scripts` section to include `lint` and `lint-schemas` commands.

I'm using **ESLint** for the linting of all the JavaScript files. I've seen this being implemented in the `website` repository. Is it okay here?

Please have a look and provide your feedback on this. Thank you. :pray:

package.json
```diff
@@ -37,10 +38,13 @@
...
   "homepage": "https://github.com/asyncapi/spec-json-schemas#readme",
   "devDependencies": {
+    "eslint": "^8.54.0",
     "mocha": "^10.0.0",
     "nyc": "^15.1.0"
   },
   "dependencies": {
-    "@types/json-schema": "^7.0.11"
+    "@types/json-schema": "^7.0.11",
+    "ajv": "^8.12.0",
```
smoya, line 47: Actually, those ajv dependencies could go un der devDependencies as you won't need it in the final npm package but only in development envs.

scripts/validate-schemas-final.js
```diff
@@ -0,0 +1,104 @@
...
+     const filePath = path.join(directoryPath, file);
+ 
+     try {
+       // Read and parse the JSON schema
+       const fileContent = fs.readFileSync(filePath, 'utf8');
+       const obj = JSON.parse(fileContent);
+ 
+       // Remove unnecessary definitions
+       delete obj.definitions['http://json-schema.org/draft-04/schema'];
+       delete obj.definitions['http://json-schema.org/draft-07/schema'];
```
smoya, line 83: I believe you said this wasn't necessary when using the `ajv.validateSchema` right? What happens when we remove those two lines?

scripts/validate-schemas-final.js
```diff
@@ -0,0 +1,104 @@
...
+ 
+       // Compile the schema
```
smoya, line 85: Nitpick:
```suggestion
       // Validate the schema
```

scripts/validate-schemas-final.js
```diff
@@ -0,0 +1,104 @@
+const fs = require('fs');
+const path = require('path');
+
+
+function validationDraft04(draft, startFileName, endFileName){
```
smoya, line 5: This function seems a literal copy of the `validation` one. Could we just only use `validation` and remove this one?

scripts/validate-schemas-final.js
```diff
@@ -0,0 +1,104 @@
...
+     } catch (error) {
+       console.error(`${file}: Error reading or parsing JSON Schema:`, error.message);
+       process.exit(1);
+     }
+   });
+}
+
+
+validationDraft04('draft04', '1.0.0-without-$id.json', '1.2.0.json');
+validation('draft07', '2.0.0-rc1.json', ['README.md', 'all.schema-store.json']);
```
smoya, line 104: I' a bit concerned about the fact you rely on alphabetical order for picking up the schemas to validate. I think this strategy can create several problems in the future, problems that could be hidden and ending up in schemas not being validated because they fall alphabetically after, in this case, `all.schema-store.json`. 

Do you think we can find a better alternative to this?
As a matter of fact, I can tell you no more draft-4 is gonna be used besides those few files, and draft7 is gonna be used in long term so we can consider it as the default.

package-lock.json
```diff
@@ -1,25 +1,52 @@
 {
   "name": "@asyncapi/specs",
   "version": "5.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 3,
```
smoya, line 4: Just letting you be aware this is changing from lockfileVersion 1 to 3. We are covered by CI so I would say there is nothing to be concerned about.

.eslintrc.json
```diff
@@ -0,0 +1,12 @@
+{
```
smoya, line 1: In other repositories, we do use `.eslintrc` file like https://github.com/asyncapi/parser-js/blob/master/.eslintrc.
I would say we can copy from there and remove Typescript related things, among others. 

WDYT?

.github/workflows/if-nodejs-pr-testing.yml
```diff
@@ -76,3 +76,6 @@ jobs:
       - if: steps.packagejson.outputs.exists == 'true'
         name: Run release assets generation to make sure PR does not break it
         run: npm run generate:assets --if-present
+      - if: steps.packagejson.outputs.exists == 'true'
+        name: Run JSON Schema Validator
+        run: npm run lint-schemas --if-present
```
smoya, line 81: no `--if-present` needed here
```suggestion
        run: npm run lint-schemas
```

package.json
```diff
@@ -11,7 +11,8 @@
     "prepublishOnly": "npm run build",
     "bundle": "cd tools/bundler && npm i && npm run bundle",
     "startNewVersion": "newVersion=$npm_config_new_version node scripts/add-new-version.js",
-    "lint": "echo 'No linter integrated yet'",
+    "lint": "npx eslint . --ext .js",
```
smoya, line 14: You can borrow this line from Parser, for example: https://github.com/asyncapi/parser-js/blob/master/package.json#L36

Also feel free (optional) to add the `lint:fix` lines as well.

scripts/validate-schemas-final.js
char0n, line 104: Just an idea here: would it be more future proof to have a script traversing all the files in `schema/` directory, reading every file `$shema` to determine it's version (Draft) and then validating according to that determined version? Script designed like that wouldn't have to be modified if new version of the AsyncAPI spec will land. This feature proof script must also accept list of ignored files for it to work properly.

smoya (changes requested):
I'm so happy to see how close we are of having validation of our JSON Schema files in CI!

You did a really great job here 👏 👏 .

I wrote some comments, suggestions, and other things that I believe we must fix. Feel free to ask any question.

package.json
char0n, line 14: Looking at https://github.com/asyncapi/spec-json-schemas/issues/436, was the intention there to add JavaScript linting as well? After reading the issue I'm still not sure.

package.json
AnimeshKumar923: Understood.
resolved with da96b36

scripts/validate-schemas-final.js
AnimeshKumar923: let me try running the script without including these lines

AnimeshKumar923: I have some observations I would like to report regarding this:

- If I remove the above said lines from the `validationDraft04` function, it gives no error.
- But if I remove that from the `validation` function (which validates the draft-07 files), it gives the following error :point_down: 
```
1.0.0-without-$id.json: JSON Schema is valid!
1.0.0.json: JSON Schema is valid!
1.1.0-without-$id.json: JSON Schema is valid!
1.1.0.json: JSON Schema is valid!
1.2.0-without-$id.json: JSON Schema is valid!
1.2.0.json: JSON Schema is valid!
2.0.0-rc1.json: JSON Schema is not valid: [
  {
    instancePath: '/definitions/http:~1~1json-schema.org~1draft-04~1schema/properties/multipleOf/exclusiveMinimum',
    schemaPath: '#/properties/exclusiveMinimum/type',
    keyword: 'type',
    params: { type: 'number' },
    message: 'must be number'
  }
]
```

If it allows, I think we should keep those lines for the sake of not breaking the overall validation process :sweat_smile: Thoughts? :eyes:

scripts/validate-schemas-final.js
AnimeshKumar923: Thanks!

resolved with a71f753

scripts/validate-schemas-final.js
AnimeshKumar923: I divided into two function because of the difficulties faced between `draft-04` and `draft-07` validation.

I read [this line](https://ajv.js.org/json-schema.html#draft-04) from the ajv documentation for `draft-04` :point_right:   **`You can only use this import with JSON Schema draft-04, you cannot combine multiple JSON Schema versions in this ajv instance.`**

Since, we can only use one instance at once, I thought of making two function to validate the two different draft versions.

The `validationDraft04` starts from `1.0.0-without-$id.json` and ends on `1.2.0.json` because files ranging between these (and including those two) are of `draft-04` (I searched it in vscode). 

Hence, I made a separate function to validate `draft-04` files and since the new schemas from `2.0.0.json` are using `draft07`, I assumed future schemas will be written using `draft-7` or more latest draft versions.

For the `validation` function, it's range starts from `2.0.0-rc1.json` and it doesn't have a defined end point so if any new schema is being added, that too will be validated. It is having two exceptions `README.md` and `all.schema-store.json` namely.

So, if we don't want to validate the earlier schemas and only validate new incoming version or assuming newly developed schemas will have `draft-07` in them, we can remove the `validationDraft04` function.

WDYT?

package.json
smoya: No, it wasn't. And I think it will be better to remove it from this PR and rather add it later on. The only requirement I would say we should ask for in any PR is to, at least, a basic linting by eslint on each code pushed. Even though running it locally.

AnimeshKumar923: so should I remove the JavaScript linting from this PR?
open a new issue and a follow-up PR for that issue adding the JS linting?

scripts/validate-schemas-final.js
AnimeshKumar923: > I' a bit concerned about the fact you rely on alphabetical order for picking up the schemas to validate. I think this strategy can create several problems in the future, problems that could be hidden and ending up in schemas not being validated because they fall alphabetically after, in this case, all.schema-store.json.

okay, I understand your point.

> Do you think we can find a better alternative to this?

Let me think and I'll suggest something

> As a matter of fact, I can tell you no more draft-4 is gonna be used besides those few files, and draft7 is gonna be used in long term so we can consider it as the default.

Okay, that's a bit of relief for me to hear :sweat_smile:

AnimeshKumar923: >  think this strategy can create several problems in the future, problems that could be hidden and ending up in schemas not being validated because they fall alphabetically after, in this case, all.schema-store.json.

could you please provide me some cases which are supposed to be failed?

AnimeshKumar923: I tried running the script for the `all.schema-store.json` and it gives the following output :point_down: 

```
all.schema-store.json: Error reading or parsing JSON Schema: Cannot convert undefined or null to object
```

AnimeshKumar923: > think this strategy can create several problems in the future, problems that could be hidden and ending up in schemas not being validated because they fall alphabetically after, in this case, all.schema-store.json

I modified the script a bit for the `validation` function :point_down: 
```js
function validation (excludedFiles){

  const Ajv = require('ajv');
  const ajv = new Ajv();

  const directoryPath = './schemas';


  const files = fs.readdirSync(directoryPath);

  const filteredFiles = files.filter(file => !excludedFiles.includes(file));


  // Iterate through the filtered files
  filteredFiles.forEach(file => {
    // Construct the full path to the JSON schema file
    const filePath = path.join(directoryPath, file);

    try {
      // Read and parse the JSON schema
      const fileContent = fs.readFileSync(filePath, 'utf8');
      const obj = JSON.parse(fileContent);

      // Remove unnecessary definitions
      delete obj.definitions['http://json-schema.org/draft-04/schema'];
      delete obj.definitions['http://json-schema.org/draft-07/schema'];

      // Validate the schema
      const validate = ajv.validateSchema(obj);

      // Check if the schema is valid
      if (validate) {
        console.log(`${file}: JSON Schema is valid!`);
      } else {
        console.error(`${file}: JSON Schema is not valid:`, ajv.errors);
        process.exit(1);
      }
    } catch (error) {
      console.error(`${file}: Error reading or parsing JSON Schema:`, error.message);
      process.exit(1);
    }
  });
}


const excludeFiles=['README.md', 'all.schema-store.json', '1.0.0-without-$id.json', '1.0.0.json', '1.1.0-without-$id.json','1.1.0.json', '1.2.0-without-$id.json', '1.2.0.json', '2.0.0-rc1-without-$id.json'];

validationDraft04('draft04', '1.0.0-without-$id.json', '1.2.0.json');

validation(excludeFiles);
```
made a `excludeFiles` array containing the files to exclude which are `README.md`, `all.schema-store.json` and the draft-4 files. Passed it to the `validation` function and made a check inside the function to read all files except these files.

For the cases which I could think of as per my current level of knowledge, I created 2 files named `0.0.0.json` and `.es.json` for the sake of testing. Executed the script for these files and both these files were read during the execution.

These are the output for the updated script :point_down: 
```console
1.0.0-without-$id.json: JSON Schema is valid!
1.0.0.json: JSON Schema is valid!
1.1.0-without-$id.json: JSON Schema is valid!
1.1.0.json: JSON Schema is valid!
1.2.0-without-$id.json: JSON Schema is valid!
1.2.0.json: JSON Schema is valid!
.es.json: JSON Schema is valid!
0.0.0.json: JSON Schema is valid!
2.0.0-rc1.json: JSON Schema is valid!
2.0.0-rc2-without-$id.json: JSON Schema is valid!
2.0.0-rc2.json: JSON Schema is valid!
2.0.0-without-$id.json: JSON Schema is valid!
2.0.0.json: JSON Schema is valid!
2.1.0-without-$id.json: JSON Schema is valid!
2.1.0.json: JSON Schema is valid!
2.2.0-without-$id.json: JSON Schema is valid!
2.2.0.json: JSON Schema is valid!
2.3.0-without-$id.json: JSON Schema is valid!
2.3.0.json: JSON Schema is valid!
2.4.0-without-$id.json: JSON Schema is valid!
2.4.0.json: JSON Schema is valid!
2.5.0-without-$id.json: JSON Schema is valid!
2.5.0.json: JSON Schema is valid!
2.6.0-without-$id.json: JSON Schema is valid!
2.6.0.json: JSON Schema is valid!
```
Those two files are also being read along with the other draft-7 files. Is this correct? Or am I missing something? :eyes:

package-lock.json
AnimeshKumar923: > Just letting you be aware this is changing from lockfileVersion 1 to 3.

Okay, I see...

> We are covered by CI so I would say there is nothing to be concerned about.

That's good to know! :smile:

.github/workflows/if-nodejs-pr-testing.yml
AnimeshKumar923: resolved with 8399eb9

package.json
AnimeshKumar923: > No, it wasn't. And I think it will be better to remove it from this PR and rather add it later on. The only requirement I would say we should ask for in any PR is to, at least, a basic linting by eslint on each code pushed. Even though running it locally.
_Originally posted by @smoya in https://github.com/asyncapi/spec-json-schemas/pull/452#discussion_r1402729934_

Do we need this here after [your recent comment](https://github.com/asyncapi/spec-json-schemas/pull/452#discussion_r1402729934) about including the JavaScript linting?

.eslintrc.json
AnimeshKumar923: > No, it wasn't. And I think it will be better to remove it from this PR and rather add it later on. The only requirement I would say we should ask for in any PR is to, at least, a basic linting by eslint on each code pushed. Even though running it locally.
_Originally posted by @smoya in https://github.com/asyncapi/spec-json-schemas/pull/452#discussion_r1402729934_

Do we need this here after [your recent comment](https://github.com/asyncapi/spec-json-schemas/pull/452#discussion_r1402729934) about including the JavaScript linting?

AnimeshKumar923:
Hey @smoya,

One more thing I wanted to tell that I've deleted the `2.0.0-rc1-without-$id.json` file because it was repeatedly failing the validation script.
Should I add it back?

The error it is giving is this :point_down: 
```
2.0.0-rc1-without-$id.json: JSON Schema is not valid: [
  {
    instancePath: '/definitions/json-schema-draft-07-schema/properties/multipleOf/exclusiveMinimum',
    schemaPath: '#/properties/exclusiveMinimum/type',
    keyword: 'type',
    params: { type: 'number' },
    message: 'must be number'
  }
]
```

AnimeshKumar923:
> I'm so happy to see how close we are of having validation of our JSON Schema files in CI!

I'm also happy and surprised myself that I've gotten this far! I was not having any knowledge of JavaScript or Schemas when I was starting this issue. I'm like...just WOW! That's some progress I see there... :eyes: :grin: 

> You did a really great job here 👏 👏 .

Your praise means a lot to me. I'm happy to be mentored by a Senior Engineer of your caliber who is this much knowledgeable. Really learnt a lot from you. Thank you! :bow: 

> I wrote some comments, suggestions, and other things that I believe we must fix. Feel free to ask any question.

Yes, I've responded to most of them I think. Please have a look when time permits :smiley:

scripts/validate-schemas-final.js
smoya: I don't really think that error is related with you removing those lines. Are you sure that's the reason?

scripts/validate-schemas-final.js
smoya: > Since, we can only use one instance at once, I thought of making two function to validate the two different draft versions.

But you are already instantiating one or another ajv based on the draft. See https://github.com/asyncapi/spec-json-schemas/pull/452/files#diff-1ee84b0b752624e8218489fead8e59e7dd93c1841a0b80418ecfa6cde1d57497R7

> The validationDraft04 starts from 1.0.0-without-$id.json and ends on 1.2.0.json because files ranging between these (and including those two) are of draft-04 (I searched it in vscode).

I agree @char0n's [comment](https://github.com/asyncapi/spec-json-schemas/pull/452#discussion_r1401755870). You can iterate over all schemas in `/schemas` and determine their metaschema (draft4 or draft7). You can check [the `$schema` field](https://github.com/asyncapi/spec-json-schemas/blob/master/schemas/1.0.0.json#L3) once you parse the file and you will get either `http://json-schema.org/draft-04/schema` or `http://json-schema.org/draft-07/schema`. Then you instantiate one or another ajv version. Does it sound reasonable?

scripts/validate-schemas-final.js
smoya: Please read https://github.com/asyncapi/spec-json-schemas/pull/452#discussion_r1403103955

package.json
smoya: You can remove all linting scripts. But make sure you run a minimum lint to your code before last push ok?

smoya: > so should I remove the JavaScript linting from this PR? open a new issue and a follow-up PR for that issue adding the JS linting?

Sounds like a plan

smoya:
> Hey @smoya,
> 
> One more thing I want to tell that I've deleted the `2.0.0-rc1-without-$id.json` file because it was repeatedly failing the validation script. Should I add it back?

Please read my comment on https://github.com/asyncapi/spec-json-schemas/pull/452#discussion_r1403105547

scripts/validate-schemas-final.js
AnimeshKumar923: > BTW, no file exclusion for json files under /schema should happen at a very first glance.

Got it!

scripts/validate-schemas-final.js
AnimeshKumar923: > You can iterate over all schemas in /schemas and determine their metaschema (draft4 or draft7). You can check [the $schema field](https://github.com/asyncapi/spec-json-schemas/blob/master/schemas/1.0.0.json#L3) once you parse the file and you will get either http://json-schema.org/draft-04/schema or http://json-schema.org/draft-07/schema. Then you instantiate one or another ajv version.

Okay, that will be a more optimized approach I think.

> Does it sound reasonable?

Yes, it does! I'll work on this approach

cc: @char0n Sorry, I missed [your comment above](https://github.com/asyncapi/spec-json-schemas/pull/452#discussion_r1401755870). The idea you've suggested would be more appropriate to implement. Thank you for your idea! :pray: 

***
> PS; _Sorry for the delayed response, running a bit hectic due to college exams :smiling_face_with_tear:_

scripts/validate-schemas-final.js
AnimeshKumar923: Thank you for this idea. Will work upon it! 🫡

package.json
AnimeshKumar923: > Sounds like a plan

Added it to my To-do list. Let's do this next :rocket:

AnimeshKumar923:
> > Hey @smoya,
> > One more thing I want to tell that I've deleted the `2.0.0-rc1-without-$id.json` file because it was repeatedly failing the validation script. Should I add it back?
> 
> Please read my comment on [#452 (comment)](https://github.com/asyncapi/spec-json-schemas/pull/452#discussion_r1403105547)

> _BTW, no file exclusion for json files under /schema should happen at a very first glance._

Yes, understood.

scripts/validate-schemas-final.js
AnimeshKumar923: > > Since, we can only use one instance at once, I thought of making two function to validate the two different draft versions.
> 
> But you are already instantiating one or another ajv based on the draft. See https://github.com/asyncapi/spec-json-schemas/pull/452/files#diff-1ee84b0b752624e8218489fead8e59e7dd93c1841a0b80418ecfa6cde1d57497R7

Oh, sorry. I missed it :sweat_smile:

package.json
AnimeshKumar923: > You can remove all linting scripts. But make sure you run a minimum lint to your code before last push ok?

something like this :point_down: 
```json
"lint-schemas": "npx eslint ./scripts/validate-schemas-final.js && node scripts/validate-schemas-final.js"
```

AnimeshKumar923: > Also feel free (optional) to add the lint:fix lines as well.

should I add this now to `package.json`?

smoya: No. Do not add anything about linting in package.json. Do not include the eslint dependency either. 
Just run the linter locally every time you do push your code so you ensure your code is linted.

AnimeshKumar923: >  Just run the linter locally every time you do push your code so you ensure your code is linted.

push as in?

AnimeshKumar923: > No. Do not add anything about linting in package.json. Do not include the eslint dependency either

understood.

AnimeshKumar923: marking as resolved because it will converted into a new issue in the future...

AnimeshKumar923:
@smoya 

I have updated the code to match the suggestions provided by you and @char0n.
There are some things I would like to report:

***

### Case 1
- This schema :point_right: `2.0.0-rc1-without-$id.json`, it is not validating when I keep these lines
```js
delete obj.definitions['http://json-schema.org/draft-04/schema'];
delete obj.definitions['http://json-schema.org/draft-07/schema'];
```
And I can't find the reason for this. So, I've kept this in the `excludedFiles` block as of now. The error it is giving is below :point_down: 
```
2.0.0-rc1-without-$id.json: JSON Schema is not valid: [
  {
    instancePath: '/definitions/json-schema-draft-07-schema/properties/multipleOf/exclusiveMinimum',
    schemaPath: '#/properties/exclusiveMinimum/type',
    keyword: 'type',
    params: { type: 'number' },
    message: 'must be number'
  }
]
```

***
### Case 2
- If I comment out this line `delete obj.definitions['http://json-schema.org/draft-07/schema']` and keep this `delete obj.definitions['http://json-schema.org/draft-04/schema']`, the scripts are validating correctly.
- If I comment out this line `delete obj.definitions['http://json-schema.org/draft-04/schema']`and keep this `delete obj.definitions['http://json-schema.org/draft-07/schema']`, it is giving error for the file: `2.0.0-rc1.json`. The error :point_down: 
```
2.0.0-rc1.json: JSON Schema is not valid: [
  {
    instancePath: '/definitions/http:~1~1json-schema.org~1draft-04~1schema/properties/multipleOf/exclusiveMinimum',
    schemaPath: '#/properties/exclusiveMinimum/type',
    keyword: 'type',
    params: { type: 'number' },
    message: 'must be number'
  }
]
```

***
### Case 3
- The `all.schema-store.json` file is giving the following error :point_down: 
```
all.schema-store.json: Error reading or parsing JSON Schema: Cannot convert undefined or null to object
```
***

Please have a look @smoya @char0n at these issues and provide your feedback.

scripts/validate-schemas-final.js
AnimeshKumar923: implemented the idea through eccb4ab

marking as resolved

scripts/validate-schemas-final.js
AnimeshKumar923: Please see the [new updated script](https://github.com/asyncapi/spec-json-schemas/pull/452/files#diff-1ee84b0b752624e8218489fead8e59e7dd93c1841a0b80418ecfa6cde1d57497), I've updated it as per your suggestion. Thank you for this! :pray: 

Marking this as resolved

scripts/validate-schemas-final.js
smoya: You can use an external JSON Schema validator in order to see if those files are valid or not.
I've test it, and both seem invalid.

- `2.0.0-rc1.json`: https://www.jsonschemavalidator.net/s/r114nPn0
- `2.0.0-rc1-without-$id.json`: https://www.jsonschemavalidator.net/s/gocJb7o3

So maybe we just discovered, thanks to your script, that those have been there in a broken state for a while.
What do you believe we should do then with those two schemas? If they are not valid, does it make sense to even keep them here?

Pinging the rest of code owners here @fmvilas @derberg @dalelane @char0n.

smoya: Coming back to your script, the weird thing is that the `2.0.0-rc1.json` seems valid in your script, meanwhile `ajv` cli says both are invalid:

```
❯ ajv compile -s './schemas/2.0.0-rc1*'
schema ./schemas/2.0.0-rc1-without-$id.json is invalid
error: schema is invalid: data/definitions/json-schema-draft-07-schema/properties/multipleOf/exclusiveMinimum must be number
schema ./schemas/2.0.0-rc1.json is invalid
error: schema is invalid: data/definitions/http:~1~1json-schema.org~1draft-04~1schema/properties/multipleOf/exclusiveMinimum must be number
```

smoya: It seems you don't need those:

```
delete obj.definitions['http://json-schema.org/draft-04/schema'];
delete obj.definitions['http://json-schema.org/draft-07/schema'];
```

It seems the way `validateSchema` works is that it really requires the schemas to be present and that It does not add any meta schema by default.

scripts/validate-schemas-final.js
```diff
@@ -0,0 +1,71 @@
...
+
+    try {
+      // Read and parse the JSON schema
+      const fileContent = fs.readFileSync(filePath, 'utf8');
+      const obj = JSON.parse(fileContent);
+
+        if(obj.$schema === 'http://json-schema.org/draft-04/schema'){
+
+            // console.log('draft-04'); // for debugging purpose
+            const Ajv = require('ajv-draft-04');
```
smoya, line 30: Since this `require` is located inside the `forEach` loop, you are creating a new AJV instance on each iteration. This is not only not needed but not good in terms of performance and resource consumption.

I suggest you to move your requires to the top of the file. You can change the const name to `AjvDraft07` and `AjvDraft04` so you can either use one or another later here.

scripts/validate-schemas-final.js
```diff
@@ -0,0 +1,71 @@
...
+            }
+        }
+    } catch (error) {
+      console.error(`${file}: Error reading or parsing JSON Schema:`, error.message);
+      process.exit(1);
+    }
+  });
+}
+
+const excludedFiles=['README.md', '2.0.0-rc1-without-$id.json'];
```
smoya, line 69: I would say we should filter any non `.json` file for default

scripts/validate-schemas-final.js
```diff
@@ -0,0 +1,71 @@
...
+            delete obj.definitions['http://json-schema.org/draft-07/schema'];
+            // Validate the schema
+            const validate = ajv.validateSchema(obj);
+
+            // Check if the schema is valid
+            if (validate) {
+                console.log(`${file}: JSON Schema is valid!`);
+            } else {
+                console.error(`${file}: JSON Schema is not valid:`, ajv.errors);
+                process.exit(1);
```
smoya, line 59: The issue of exiting here is that you are early-terminating the script once it founds the first error. In case there are more than one invalid JSON Schema file, you will only see the very first.

This can be specially annoying specially when running this script during CI, because you might be fixing one schema, and then waiting for another CI run to pass and find out that you have another invalid schema and so on.

I would suggest you collect errors in an array or whatever, and in the very end of the script you check if errors exist, and if so, print and exit(1).

package.json
```diff
@@ -12,7 +12,8 @@
     "bundle": "cd tools/bundler && npm i && npm run bundle",
     "startNewVersion": "newVersion=$npm_config_new_version node scripts/add-new-version.js",
     "lint": "echo 'No linter integrated yet'",
-    "bump:version": "npm --no-git-tag-version --allow-same-version version $VERSION"
+    "bump:version": "npm --no-git-tag-version --allow-same-version version $VERSION",
+    "lint-schemas": "node scripts/validate-schemas-final.js"
```
smoya, line 16: We discussed about this somewhere but i can't find my comment, so suggesting a rename here directly: 

```suggestion
    "validate-schemas": "node scripts/validate-schemas-final.js"
```

AnimeshKumar923: > We discussed about this somewhere but i can't find my comment, so suggesting a rename here directly:

Oh, I just remembered this after your comment. Anyways, it is more appropriate naming for this. :100: 

resolved with b48c050

scripts/validate-schemas-final.js
AnimeshKumar923: > The issue of exiting here is that you are early-terminating the script once it founds the first error. In case there are more than one invalid JSON Schema file, you will only see the very first.
>
> This can be specially annoying specially when running this script during CI, because you might be fixing one schema, and then waiting for another CI run to pass and find out that you have another invalid schema and so on.

Okay, I see...I get your point. :+1: 

> Do you believe you could change your code so you could validate all files and only then print all errors at the end of the execution?

Yes, I'll look into it. 🫡

scripts/validate-schemas-final.js
AnimeshKumar923: > I would say we should filter any non `.json` file for default

Let me try that...

scripts/validate-schemas-final.js
AnimeshKumar923: > Since this require is located inside the forEach loop, you are creating a new AJV instance on each iteration. This is not only not needed but not good in terms of performance and resource consumption.

Hmm... didn't knew this, Thank you for teaching me this! :pray: 

> I suggest you to move your requires to the top of the file. You can change the const name to AjvDraft07 and AjvDraft04 (or any other name of your preference) so you can either use one or another later here.

Will implement your suggestion... :+1:

AnimeshKumar923:
> No. Do not add anything about linting in package.json. Do not include the eslint dependency either. 
> Just run the linter locally every time you do push your code so you ensure your code is linted.
>
> _Originally posted by @smoya in https://github.com/asyncapi/spec-json-schemas/pull/452#discussion_r1405173275_


.
> Just run the linter locally every time you do push your code so you ensure your code is linted.

`push your code`, what does it mean?

scripts/validate-schemas-final.js
AnimeshKumar923: > I' a bit concerned about the fact you rely on alphabetical order for picking up the schemas to validate. I think this strategy can create several problems in the future, problems that could be hidden and ending up in schemas not being validated because they fall alphabetically after, in this case, `all.schema-store.json`.
> 
> Do you think we can find a better alternative to this? As a matter of fact, I can tell you no more draft-4 is gonna be used besides those few files, and draft7 is gonna be used in long term so we can consider it as the default.

Found an alternative for this.
Resolved with eccb4ab

scripts/validate-schemas-final.js
AnimeshKumar923: > Coming back to your script, the weird thing is that the `2.0.0-rc1.json` seems valid in your script, meanwhile `ajv` cli says both are invalid (expected):
> 
> ```
> ❯ ajv compile -s './schemas/2.0.0-rc1*'
> schema ./schemas/2.0.0-rc1-without-$id.json is invalid
> error: schema is invalid: data/definitions/json-schema-draft-07-schema/properties/multipleOf/exclusiveMinimum must be number
> schema ./schemas/2.0.0-rc1.json is invalid
> error: schema is invalid: data/definitions/http:~1~1json-schema.org~1draft-04~1schema/properties/multipleOf/exclusiveMinimum must be number
> ```

If I delete this line from the script, 
```
delete obj.definitions['http://json-schema.org/draft-04/schema']:
```
then only it says that `2.0.0-rc1.json` is invalid.

AnimeshKumar923: > It seems you don't need those:
> 
> ```
> delete obj.definitions['http://json-schema.org/draft-04/schema'];
> delete obj.definitions['http://json-schema.org/draft-07/schema'];
> ```
> 
> It seems the way `validateSchema` works is that it really requires the schemas to be present and that It does not add any meta schema by default. See:
> 
> > If schema has $schema property, then the schema with this id (that should be previously added) is used to validate passed schema.
> > Source: https://ajv.js.org/api.html#ajv-validateschema-schema-object-boolean

Okay, I see. I'll remove this.

Does it mean that it is already doing the `delete` statements inclusively that we were dong exclusively in the script?

AnimeshKumar923: > You can use an external JSON Schema validator in order to see if those files are valid or not. I've test it, and both seem invalid.
> 
>     * `2.0.0-rc1.json`: https://www.jsonschemavalidator.net/s/r114nPn0
> 
>     * `2.0.0-rc1-without-$id.json`: https://www.jsonschemavalidator.net/s/gocJb7o3
> 
> 
> So maybe we just discovered, thanks to your script, that those have been there in a broken state for a while.

It is already being helpful before fully deployed, :eyes:  happy to see that :grinning: 
 
> What do you believe we should do then with those two schemas? If they are not valid, does it make sense to even keep them here?

I think if they are not being implemented anywhere, it would be best to delete them.

I see `rc-1`, `rc-2`, and then there is the `2.0.0.json` with these three having 2 files for each, one with `$id` and another `without-$id`. What is the difference between them? And which one out of those is used for real-world scenarios for v2?

scripts/validate-schemas-final.js
AnimeshKumar923: Hey Sergio!
I've updated the script for the instantiation part, please see here: 3570fe0b2cdc0d47287fad04eeab40e09ba7150c

package.json
AnimeshKumar923: marking resolved as we don't have to include linting in this PR, we will do it separately in a new issue.

smoya:
> `push your code`, what does it mean?

Push your code to github. Literal `git push`.

AnimeshKumar923:
> > `push your code`, what does it mean?
> 
> Push your code to github. Literal `git push`.

Oh, okay. I thought we need to implement this somewhere in the script. :sweat_smile:

AnimeshKumar923:
@smoya 
image

Will this be updated from your side? 
from `lint-schemas` to `validate-schemas`

smoya:
> @smoya 
> image
> 
> Will this be updated from your side? 
> from `lint-schemas` to `validate-schemas`

You still need to change the call you do in the github action

AnimeshKumar923:
> > @smoya
> > image
> > Will this be updated from your side?
> > from `lint-schemas` to `validate-schemas`
> 
> You still need to change the call you do in the github action

Yup, just did that!

scripts/validate-schemas-final.js
AnimeshKumar923: Resolved with 3570fe0

scripts/validate-schemas-final.js
AnimeshKumar923: implemented through df356456485cbe38d9eda548cc3fd758f54d1801

marking as resolved

scripts/validate-schemas-final.js
AnimeshKumar923: image
I guess this is what you suggested... :eyes: 

I've updated the code according to this. Please have a look
Ref: 17fcffdf3bab90fc807a92808eb0173467a0a209

smoya:
Ok @AnimeshKumar923, I believe all requirements are now fulfilled. Congrats for the nice job you did here! 🎉 🚀 👏

There is only one pending thing regarding the failing JSON Schema files `2.0.0-rc1.json` and `2.0.0-rc1-without-$id.json`. Since this is something that shouldn't block this PR to move forward, I think the best would be to exclude those two files from the validation. At least temporary until we decide what we do with it.

Do you believe you can add that exclusion somewhere? I'm gonna create a follow up issue for taking a decision on what to do with those schemas (which may be just affected by https://github.com/asyncapi/spec-json-schemas/issues/454).

AnimeshKumar923:
> > Do you believe you can add that exclusion somewhere?
>
> Yes, gladly!
> 'm thinking of adding excludeFiles function used earlier to pass these two files.

Hey @smoya :wave: 
Please checkout b819893 and provide your feedback. I've implemented your suggestions.

AnimeshKumar923:
@smoya bringing your attention to this issue

Please a have look at b819893 for the latest changes

scripts/validate-schemas-final.js
```diff
@@ -0,0 +1,90 @@
...
+    
+            // Check if the schema is not valid and collect errors
+            if (!validate) {
+                validationErrors.push({
+                    file,
+                    errors: obj.$schema === 'http://json-schema.org/draft-04/schema'
+                    ? ajvDraft04.errors
+                    : ajv.errors
+                });
+                // console.error(`${file}: JSON Schema is not valid:`, ajv.errors);
```
smoya, line 60: Residual line. Can be deleted.
```suggestion
                // console.error(`${file}: JSON Schema is not valid:`, ajv.errors);
```

scripts/validate-schemas-final.js
```diff
@@ -0,0 +1,90 @@
...
+            process.exit(1);
+        }
+        
+    } catch (error) {
+        console.error('\nError during validation:', error.message);
+    }
+}
+
+
+const excludedFiles=['2.0.0-rc1.json', '2.0.0-rc1-without-$id.json']; // added temporarily to avoid validation failure due to these two files. The schemas version are incorrect in these and needs to be fixed.
```
smoya, line 86: Nice 👍 👍

scripts/validate-schemas-final.js
AnimeshKumar923: Will do.

scripts/validate-schemas-final.js
AnimeshKumar923: Thank you! :bow:

package-lock.json
```diff
@@ -4691,4 +4691,4 @@
       "dev": true
     }
   }
-}
+}
```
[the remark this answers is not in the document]
AnimeshKumar923: @smoya 
The problem was that the packages were not fully installed due to fluctuating internet on my side. Please check 75c0da2, I've updated it.

I apologize for the issues :sweat:

smoya (dismissed):
LGTM!  🚀🌔

I'm so happy with the final result @AnimeshKumar923. Good job!

smoya:
@derberg you might wanna do a quick review on this one so we can merge? 🙏

package-lock.json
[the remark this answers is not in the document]
AnimeshKumar923: Resolved with 75c0da2

aeworxet:
Please fix the code smell (whoever's it is.)
https://sonarcloud.io/project/issues?id=asyncapi_asyncapi-node&pullRequest=452&resolved=false&types=CODE_SMELL

scripts/validate-schemas-final.js
smoya: Follow up issue was created https://github.com/asyncapi/spec-json-schemas/issues/455

smoya (dismissed):

smoya:
@AnimeshKumar923 there is one code smell SonarCloud complains https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=452&id=asyncapi_asyncapi-node&open=AYwW7h_-2LUoXdDhnTTh.

It is about simplifying the code basically. I agree, it can be simplified and improved and I invite you to do it, but I would say this can be done in a separate PR. To be honest, I find this code to be readable and simple enough at this stage IMHO. I would say you can create a follow up issue in order to simplify that.

Regarding linting, and as we do not have a linter in place, I won't required you to pass any linting. Instead, we will fix those once https://github.com/asyncapi/spec-json-schemas/issues/467 is done.

smoya:
@derberg ping 🙌

derberg (commented):
you cannot modify `.github/workflows/if-nodejs-pr-testing.yml` as it comes from `.github` and whatever you modify now can be overwritten later

better just modify `test` script in package.json and make it fun the `validate-schemas` script, so instead of `"test": "npm run build && nyc mocha",` do `"test": "npm run build && nyc mocha && npm run validate-schemas",`. This way you do not have to modify `.github/workflows/if-nodejs-pr-testing.yml`

last but not least, please do not introduce scripts names with `-` as we follow approach with `:`. So instead of `validate-schemas` better make `validate:schemas`


---

also did you check if you have proper error handling in script that will fail GitHub action in case of validation fails? so we know that it actually gets red when error happens? I see you have one try/catch that is just doing `console.error` and it will for sure not terminate script exacution

AnimeshKumar923:
> @AnimeshKumar923 there is one code smell SonarCloud complains https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=452&id=asyncapi_asyncapi-node&open=AYwW7h_-2LUoXdDhnTTh.
> 
> It is about simplifying the code basically. I agree, it can be simplified and improved and I invite you to do it, but I would say this can be done in a separate PR. 

It's a new thing for me (the code smell part), still learning from it.

> To be honest, I find this code to be readable and simple enough at this stage IMHO. 

Okay, I see

> I would say you can create a follow up issue in order to simplify that.

Yes, I agree. For the code optimization, I would like to do it in a new issue and it might take some time as per my current knowledge about JavaScript and programming in general. :sweat_smile: 

> Regarding linting, and as we do not have a linter in place, I won't required you to pass any linting. Instead, we will fix those once #467 is done.

Understood.

AnimeshKumar923:
Hi @derberg! :wave: 

Thank you for your detailed review. I'll definitely look into the things you mentioned. :+1:

AnimeshKumar923:
Hey @smoya @derberg :wave: 

Please have a look at the files now. I've made changes suggested [here](https://github.com/asyncapi/spec-json-schemas/pull/452#pullrequestreview-1775045082) through aeada06

derberg:
lgtm, I see also that validation runs: https://github.com/asyncapi/spec-json-schemas/actions/runs/7180191694/job/19557877973?pr=452

did you at some point of time try to simulate an error to see if test will fail?

AnimeshKumar923:
> lgtm, I see also that validation runs: https://github.com/asyncapi/spec-json-schemas/actions/runs/7180191694/job/19557877973?pr=452
> 
> did you at some point of time try to simulate an error to see if test will fail?

@derberg Yes. I created a dummy `0.0.0.json` and tested it.

Contents of `0.0.0.json` :point_down: (thanks goes to Sergio for providing this template for testing!)
```json
{
"$id":"http://foo.bar",
"$schema":"http://json-schema.org/draft-07/schema",
    "definitions":{
        "http://json-schema.org/draft-07/schema":{},
        "person":{
            "type":"object",
            "properties":{
                "name":{
                    "type":"invalid"
                }
            }
        }
    }
}
```

These are the results after running `node scripts/validate-schemas.js` :point_down: 
```console
1.0.0-without-$id.json: JSON Schema is valid!

1.0.0.json: JSON Schema is valid!

1.1.0-without-$id.json: JSON Schema is valid!

1.1.0.json: JSON Schema is valid!

1.2.0-without-$id.json: JSON Schema is valid!

1.2.0.json: JSON Schema is valid!

2.0.0-rc2-without-$id.json: JSON Schema is valid!

2.0.0-rc2.json: JSON Schema is valid!

2.0.0-without-$id.json: JSON Schema is valid!

2.0.0.json: JSON Schema is valid!

2.1.0-without-$id.json: JSON Schema is valid!

2.1.0.json: JSON Schema is valid!

2.2.0-without-$id.json: JSON Schema is valid!

2.2.0.json: JSON Schema is valid!

2.3.0-without-$id.json: JSON Schema is valid!

2.3.0.json: JSON Schema is valid!

2.4.0-without-$id.json: JSON Schema is valid!

2.4.0.json: JSON Schema is valid!

2.5.0-without-$id.json: JSON Schema is valid!

2.5.0.json: JSON Schema is valid!

2.6.0-without-$id.json: JSON Schema is valid!

2.6.0.json: JSON Schema is valid!

3.0.0-without-$id.json: JSON Schema is valid!

3.0.0.json: JSON Schema is valid!

all.schema-store.json: JSON Schema is valid!

0.0.0.json: JSON Schema is not valid: [
  {
    instancePath: '/definitions/person/properties/name/type',
    schemaPath: '#/definitions/simpleTypes/enum',
    keyword: 'enum',
    params: { allowedValues: [Array] },
    message: 'must be equal to one of the allowed values'
  },
  {
    instancePath: '/definitions/person/properties/name/type',
    schemaPath: '#/properties/type/anyOf/1/type',
    keyword: 'type',
    params: { type: 'array' },
    message: 'must be array'
  },
  {
    instancePath: '/definitions/person/properties/name/type',
    schemaPath: '#/properties/type/anyOf',
    keyword: 'anyOf',
    params: {},
    message: 'must match a schema in anyOf'
  }
]
```

and when executed `echo $?`, it resulted in `1`.

Do we have any ways of testing this dummy in the PR itself for better assurance? :eyes:

derberg (approved):
@AnimeshKumar923 this PR is already running your test script 😄 

@smoya LGTM so if it comes to me, we can merge

smoya (approved):

AnimeshKumar923:
@smoya Thank you so much for your guidance throughout this PR! I'm very happy how it turned out in the end. :bow: :rocket: :partying_face: 
_Also, let me see what can I do about the code smell in a new issue :sweat_smile:_

@derberg Thank you for your in-depth review. It was very crucial for this PR. :bow: :+1: 

Looking forward to do more contributions here.

Cheers! :clinking_glasses:

aeworxet:
@asyncapi/bounty_team

PR #458, opened 2023-11-30
docs: update readme with new release strategy

derberg:
adding one missing thing to https://github.com/asyncapi/spec-json-schemas/pull/421

so yeah, I noticed @jonaslagoni already prepared migration guide that describes new strategy -> https://github.com/asyncapi/spec-json-schemas/pull/421/files#diff-f56fedd975b2d5e7e325bfbf9d75a3a1b6483a143e59ac341e9f1d1d9144a2f1

this PR just makes sure we have proper info in readme.

for reviewers:
in readme it looks like a lot of changes but only because I refactored to get rid of too many bullet points, so the only new section is really

```
### Releases and pre-releases

This repository contains JSON Schema files for official AsyncAPI releases and also for release candidates. Before you decide to use a specific JSON Schema file in production, make sure a corresponding [official release of AsyncAPI specification](https://github.com/asyncapi/spec/releases) is produced, not a release candidate.

JSON Schema which describes a version of AsyncAPI specification that is not yet officially released is considered an unstable pre-release that can change anytime and is not considered to be a breaking-change.

If you want to make sure you only use stable schemas, you have to make sure that you use only certain schema versions, not all by default.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,24 +1,39 @@
 ![npm](https://img.shields.io/npm/v/@asyncapi/specs?style=for-the-badge) ![npm](https://img.shields.io/npm/dt/@asyncapi/specs?style=for-the-badge)
 
-# AsyncAPI
+## Overview
 
 This is a mono repository, which provides all the JSON Schema documents for validating AsyncAPI documents.
 
-## Overview
+### Two types of schemas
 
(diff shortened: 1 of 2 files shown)


fmvilas (approved):
LGTM 👍

PR #462, opened 2023-11-30
feat!: add 3.0.0 validation support

jonaslagoni:
When this is merged, from version 6 and onwards, all pre-release spec versions will be released as a regular feature request, before the spec itself is released.

The pre-release version will be released as if it was not a pre-release, for example for AsyncAPI 3.0, it will be released as normal:
```js
module.exports = {
    'schemas': {
        ...
        '3.0.0': require('./schemas/3.0.0.json'),
    },
    'schemasWithoutId': {
        ...
        '3.0.0': require('./schemas/3.0.0-without-$id.json'),
    }
};
```

However, while the it's still a pre-release, the underlying schemas CAN contain breaking changes from version to version, up until the spec is released. This means that one AsyncAPI document using v3 in the pre-release stage might be valid in `6.0.0`, but invalid in the `6.1.0`. This ONLY applies to pre-release schemas, and NOT regular ones that is set in stone.

If you want to make sure you don't use a schema not released yet, you have to whitelist which versions you allow in your tool.


jonaslagoni:
This PR is a duplicate of https://github.com/asyncapi/spec-json-schemas/pull/208 as it was incorrectly merged.

derberg (approved):

PR #470, opened 2023-12-13
fix: trigger release

smoya:
**Description**

Due to an issue with semantic-release not understanding a force push we had to do in order to fix authors of [a commit that was already released](https://github.com/asyncapi/spec-json-schemas/commit/fdfa22480ea95ac8e9c33e4f49538ca761eb7f15) in v6.0.1, we had to re-tag that version in order to point to the new commit hash. That triggered a release wrongly because the race condition between me manually creating the tag and CI running semantic-release.

This PR just commits an empty commit that will, if everything goes as expected, trigger a new minor release (according to release messages after v6.0.1)
 
cc @derberg


derberg (approved):
lgtm,
changed title as `ci:` would not trigger release

PR #481, opened 2024-01-08
feat: add new kafka topic configuration properties

gokerakc:
**Description**
The following updates are introduced for the kafka topic configuration object:

- Updated additionalProperties as true.
- Added 4 new properties as proposed in the issue.

Please check the related [issue](https://github.com/asyncapi/bindings/issues/231) for more details.

Note: Once we are happy with the changes I'll create another PR to update the bindings' README file in the other repo.

**Related issue(s)**
Resolves the following issue: https://github.com/asyncapi/bindings/issues/231


dpwdec:
Very much needed! 👍 We're struggling to work around these gaps with `x-` extensions.

bindings/kafka/0.5.0/channel.json
```diff
@@ -0,0 +1,96 @@
...
+          "description": "The [`delete.retention.ms`](https://kafka.apache.org/documentation/#topicconfigs_delete.retention.ms) configuration option.",
+          "type": "integer",
+          "minimum": 0
+        },
+        "max.message.bytes": {
+          "description": "The [`max.message.bytes`](https://kafka.apache.org/documentation/#topicconfigs_max.message.bytes) configuration option.",
+          "type": "integer",
+          "minimum": 0
+        },
+        "schemaValidationForKeyEnabled": {
```
dalelane, line 61: Is this referring to the Confluent schema validation feature? https://docs.confluent.io/platform/current/schema-registry/schema-validation.html

bindings/kafka/0.5.0/channel.json
```diff
@@ -0,0 +1,96 @@
...
+          "description": "It shows whether the schema validation for the message key is enabled.",
+          "type": "boolean"
+        },
+        "schemaLookupStrategyForKey": {
+          "description": "The name of the schema lookup strategy for the message key.",
+          "type": "string"
+        },
+        "schemaValidationForValueEnabled": {
```
dalelane, line 69: As above - what does this relate to? Is this the Confluent Cloud schema validation feature? https://docs.confluent.io/platform/current/schema-registry/schema-validation.html#

bindings/kafka/0.5.0/channel.json
```diff
@@ -0,0 +1,96 @@
...
+          "description": "It shows whether the schema validation for the message value is enabled.",
+          "type": "boolean"
+        },
+        "schemaLookupStrategyForValue": {
```
dalelane, line 73: What is your thinking about having this here compared with where it is currently in the message bindings? Do you think we need it in both places, or are you thinking we'd move it here?

bindings/kafka/0.5.0/channel.json
gokerakc: Yes, it is. I didn't check if the other vendors have a similar feature but for confluent users it's useful.

bindings/kafka/0.5.0/channel.json
gokerakc: https://github.com/asyncapi/spec-json-schemas/pull/481#discussion_r1449023289

bindings/kafka/0.5.0/channel.json
gokerakc: The current message binding object has a property called **schemaLookupStrategy** and it seems like this is the strategy for the message key.

In Confluent, you can set different strategies for your message value and your message key and because these configs exist under the [topic configurations](https://docs.confluent.io/platform/current/installation/configuration/topic-configs.html) I thought keeping those values in topicConfiguration object makes sense.

About the message binding object, I guess we can remove the **schemaLookupStrategy** property and start to use the  **schemaLookupStrategyForValue** and **schemaLookupStrategyForKey** properties added to the topic configuration to avoid having the same value in two different places.

bindings/kafka/0.5.0/channel.json
dalelane: ah I see - sorry, I should've clicked through to the issue, I just looked at the pull request in isolation - my bad

bindings/kafka/0.5.0/channel.json
gokerakc: Any thoughts about this ? @dalelane

derberg:
Hey, please just remember that JSON Schema is just a tool, not a spec itself so whatever is added to JSON Schema, first need to be added to Kafka Binding spec -> https://github.com/asyncapi/bindings/tree/master/kafka

gokerakc:
I've updated this PR as a draft because the following PR needs to be reviewed first:

- https://github.com/asyncapi/bindings/pull/238

bindings/kafka/0.5.0/channel.json
gokerakc: IMO, we can keep both of them for now because message binding is what you declare for your clients to understand about the message format and topic configuration is what you tell the broker about the channel and its attributes. If it doesn't make sense to keep both then we can remove it in the further versions.

By the way, it seems like I need to update the bindings repo first so we can continue our discussion under this PR -> https://github.com/asyncapi/bindings/pull/238.

gokerakc:
Given the related bindings PR (https://github.com/asyncapi/bindings/pull/238) got merged I think we can continue with this one.

I've added a new schema version (0.5.0) to reflect the new changes in the kafka topic configuration object.

It would be great if you could re-review this PR @dalelane.

dalelane (dismissed):
sorry for the delay in replying - I was on vacation last week

this looks good to me - thanks for the updates

gokerakc:
Sorry guys, I've messed around with my commits. I had a problem verifying my signature but now it's fixed.

It seems like extra checks (_Test NodeJS PR_ ones) were added because of my commit changes. If it's necessary, I can close this PR and create a new one.

Can you please re-approve the PR? @dalelane

dalelane (approved):

gokerakc:
Hey guys, can you please approve that 1 workflow? It seems like we need to run that workflow before the merge.  @dalelane @smoya

derberg (commented):
@gokerakc hey, I see we overlooked and did not make sure you add new version to proper definitions. Can you add `0.5.0` to:
- https://github.com/asyncapi/spec-json-schemas/blob/master/definitions/3.0.0/operationBindingsObject.json#L134
- https://github.com/asyncapi/spec-json-schemas/blob/master/definitions/3.0.0/serverBindingsObject.json#L53
- https://github.com/asyncapi/spec-json-schemas/blob/master/definitions/3.0.0/channelBindingsObject.json#L87
- https://github.com/asyncapi/spec-json-schemas/blob/master/definitions/3.0.0/messageBindingsObject.json#L134

and I do not only mean an enum, to add it to enum, but also respective `if` statement?

if you do it just ping me and I will review and merge with higher priority

as for now, new version is not really in place as anyway documents with `0.5.0` binding are not accepted as valid

dalelane:
@derberg @gokerakc sorry I missed that when I did the review - I've added a PR now at https://github.com/asyncapi/spec-json-schemas/pull/501

gokerakc:
Thanks @dalelane!

Issue #494, opened 2024-03-08
AsyncAPI v3 shows warning in Studio and IntelliJ plugin when referencing a json schema


ignacio-nyt:
### Describe the bug.

AsyncAPI v3 shows warning in Studio and IntelliJ plugin when referencing a json schema file. The actual AsyncAPI document is rendered correctly but the warning in the editor persists.

### Expected behavior

AsyncAPI v3 should be able to reference external json schema files without warnings.

### Screenshots

![image (6)](https://github.com/asyncapi/spec-json-schemas/assets/89947202/6a17d68c-2cb7-40e3-bc41-fbdff2233a53)
![image (5)](https://github.com/asyncapi/spec-json-schemas/assets/89947202/ee7b31b9-66a3-4ee6-80c5-86a83862582f)

### How to Reproduce

1. I first created a valid JSON schema
2. I then created an AsyncAPI v3 doc referencing a JSON schema 
4. Then I see the AsyncAPIv3 document rendering the JSON schema correctly but still showing warnings

Please rename the attached AsyncAPI v3 doc `test.txt` to `test.yaml`
[test.txt](https://github.com/asyncapi/spec-json-schemas/files/14541883/test.txt)
[invoice_posted.schema.json](https://github.com/asyncapi/spec-json-schemas/files/14541885/invoice_posted.schema.json)

### 🥦 Browser

None

### 👀 Have you checked for similar open issues?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue ?

None

jonaslagoni:
I think we have a problem with the JSON Schema files that describe payloads.

Payloads are defined with AnySchema: https://github.com/asyncapi/spec-json-schemas/blob/35ea79055d09bdd319ed816e4ec169a12f0bb7a5/definitions/3.0.0/messageObject.json#L19
AnySchema is defined with either schema or multiFormatSchema: https://github.com/asyncapi/spec-json-schemas/blob/master/definitions/3.0.0/anySchema.json#L8
Schema is basically just JSON Schema: https://github.com/asyncapi/spec-json-schemas/blob/35ea79055d09bdd319ed816e4ec169a12f0bb7a5/definitions/3.0.0/schema.json#L5
[...]

PR #495, opened 2024-03-10
fix: wrong validation with AsyncAPI v3 in VS Code and JetBrains IDEs

Pakisan:
Fix specification validation erros in JetBrains and VS Code

Closes https://github.com/asyncapi/spec-json-schemas/issues/494
Closes https://github.com/asyncapi/jasyncapi-idea-plugin/issues/49


jonaslagoni (changes requested):
@Pakisan you need to adapt the [`definition`](https://github.com/asyncapi/spec-json-schemas/tree/master/definitions) files, not the bundled `schemas` 🙂

Pakisan:
> @Pakisan you need to adapt the [`definition`](https://github.com/asyncapi/spec-json-schemas/tree/master/definitions) files, not the bundled `schemas` 🙂 

Yeah, expected this reply, because I saw that schemas were generated automatically

How to build schemas? I need to change only definitions?

jonaslagoni:
@Pakisan you can run the npm command [`npm run bundle`](https://github.com/asyncapi/spec-json-schemas/blob/35ea79055d09bdd319ed816e4ec169a12f0bb7a5/package.json#L12) to trigger the bundling process.

Pakisan:
> @Pakisan you can run the npm command [`npm run bundle`](https://github.com/asyncapi/spec-json-schemas/blob/35ea79055d09bdd319ed816e4ec169a12f0bb7a5/package.json#L12) to trigger the bundling process.
> 
> Yep, only need to change the definitions, and everything else will be taken care of in the release process 🙂

It may be related with references availability?

I can open https://asyncapi.com/definitions/3.0.0/Reference.json, but can't https://www.asyncapi.com/bindings/sqs/0.2.0/channel.json

```sh
➜  ~ curl https://www.asyncapi.com/bindings/sqs/0.2.0/channel.json
<!DOCTYPE html><html lang="en"><head><meta charSet="utf-8"/><meta name="viewport" content="width=device-width"/><meta name="next-head-count" content="2"/><link rel="preload" href="/_next/static/css/70fc8c1b3046a2a5.css" as="style"/><link rel="stylesheet" href="/_next/static/css/70fc8c1b3046a2a5.css" data-n-g=""/><noscript data-n-css=""></noscript><script defer="" nomodule="" src="/_next/static/chunks/polyfills-c67a75d1b6f99dc8.js"></script><script src="/_next/static/chunks/webpack-309fbebe2073f18c.js" defer=""></script><script src="/_next/static/chunks/framework-79bce4a3a540b080.js" defer=""></script><script src="/_next/static/chunks/main-5ab4e93313090fd8.js" defer=""></script><script src="/_next/static/chunks/pages/_app-e51460248b2aa259.js" defer=""></script><script src="/_next/static/chunks/pages/404-21869d41b96d4a23.js" defer=""></script><script src="/_next/static/uYuiB21UZi5IxEWUG1fEZ/_buildManifest.js" defer=""></script><script src="/_next/static/uYuiB21UZi5IxEWUG1fEZ/_ssgManifest.js" defer=""></script></head><body><div id="__next" data-reactroot=""></div><script id="__NEXT_DATA__" type="application/json">{"props":{"pageProps":{}},"page":"/404","query":{},"buildId":"uYuiB21UZi5IxEWUG1fEZ","nextExport":true,"autoExport":true,"isFallback":false,"scriptLoader":[]}</script></body></html>%
```

Because SQS channels has right reference definition:

https://github.com/asyncapi/spec-json-schemas/blob/35ea79055d09bdd319ed816e4ec169a12f0bb7a5/bindings/sqs/0.2.0/channel.json#L18-L21

which will be handled wrongly

https://github.com/asyncapi/spec-json-schemas/blob/35ea79055d09bdd319ed816e4ec169a12f0bb7a5/schemas/3.0.0-without-%24id.json#L5622-L5625

jonaslagoni:
If you are using any of the schemas in https://github.com/asyncapi/spec-json-schemas/tree/master/schemas, you don't need to go remote to fetch the schemas, its only if you work with a definition for example.

The way the $id and $ref works in these definitions is that the bundler matches them locally, the same when you try to validate with the bundled schemas, that's simply part of the JSON Schema spec, that if a resource can be found locally that will be used.

So basically, just see the references as matching together, even if the remote URL does not exist 🙂 Even though it definitely should at some point

jonaslagoni:
@smoya what do we need to change to publish the bindings publically?

smoya:
@jonaslagoni to either create a new or modify the following Netlify edge-function https://github.com/asyncapi/website/blob/master/netlify/edge-functions/serve-definitions.ts and add a new redirection rule into https://github.com/asyncapi/website/blob/master/netlify.toml

tools/bundler/index.js
```diff
@@ -165,7 +165,13 @@ function getDefinitionName(def) {
   }
   if (def.startsWith('http://asyncapi.com/bindings')) {
     const result = bindingsRegex.exec(def);
-    if (result) return `${result[1].replace('/', '-')}-${result[2]}-${result[3]}`;
+    if (result) {
```
jonaslagoni, line 168: Might need some comments here explaining what what is 😅

jonaslagoni: Guess this is where the bug is for bindings?

Pakisan: done

jonaslagoni (approved):
Nice catch and fix 👌 

As I am not codeowner someone else gotta have a look 😄

Pakisan:
@fmvilas @derberg @dalelane @smoya @char0n @GreenRover 

Sorry for massive calling, but this issue affects users in VS Code and JetBrains IDEs

I checked fix here - https://www.jsonschemavalidator.net and inside my plugin

Because our cli is thinking that everything is ok with provided schemas - https://github.com/asyncapi/jasyncapi-idea-plugin/issues/49#issuecomment-1986978991

GreenRover (dismissed):
i not understand the `if` `then` construct but the rest looks good.

dalelane (approved):

Pakisan:
@smoya @char0n hey!

If everything is ok, I want to merge it today

.gitignore
```diff
@@ -1,4 +1,11 @@
 node_modules
 .nyc_output
 .vscode
-coverage
\ No newline at end of file
+coverage
+
+.DS_Store
```
smoya, line 6: I believe this line is enough to ignore this file anywhere. No need to add the rest.

Additionally, I recommend adding those generic additions to your own global .gitignore file in`~/.gitignore`.

smoya:
> If everything is ok, I want to merge it today

TBH I wanted to take some time to review it, as I need to understand the context and the changes. But you have technically 2 +1 from owners anyway.

smoya:
Please @Pakisan hold on so I can review this PR **on Monday**. I really need to understand the implication of this change (probably will be OK, but need to understand as I worked on that particular part of the JSON Schema recently).

definitions/3.0.0/multiFormatSchema.json
```diff
@@ -54,7 +54,7 @@
         "description": "Definition of the message payload. It can be of any type but defaults to Schema Object. It MUST match the schema format defined in schemaFormat, including the encoding type. E.g., Avro should be inlined as either a YAML or JSON object instead of as a string to be parsed as YAML or JSON. Non-JSON-based schemas (e.g., Protobuf or XSD) MUST be inlined as a string."
       }
     },
-    "allOf": [
+    "anyOf": [
```
smoya, line 57: After a deep investigation, I concluded this **cannot** be changed to `anyOf` and **should** stay as `allOf`.

The reason is that `anyOf` will try to match with any schema in the array. The problem is that since those schemas are `if`, and an `if` conditional fails, it is considered an empty schema, which is equivalent to `true` assertion of that element in the `anyOf` array.

See this example I made by extracting the logic in place and reducing the complexity for test purposes: https://www.jsonschemavalidator.net/s/q0RfvVqb

As you can see, the `if` is failing (doesn't match the schemaFormat), but the `anyOf` is resulting in a true assertion. Totally against what we expect.

#### How can we fix then the issue?

Let's read what the JSON Schema validator says:

Slack_TGf24bNa@2x

The issue comes not at this line, but at [this other one](https://github.com/asyncapi/spec-json-schemas/blob/master/schemas/3.0.0-without-%24id.json#L2820-L2827).

```json
"oneOf": [
    {
        "$ref": "#/definitions/Reference"
    },
    {
        "$ref": "#/definitions/schema"
    }
]
```

The point is that the keyword `$ref` is both defined in the `Reference` schema and in the `schema`, which is a superset of JSON Schema and it does include the `$ref` keyword itself.
That's why JSON Schema validation says the schema is matching with more than one schema in this`oneOf`.

By changing that `oneOf` to `anyOf`, we ensure that by matching with any of those schemas will be considered a valid doc.

I spent some time trying to find edge-cases for the change I'm suggesting. For example, what if we change how our Reference Object (`$ref`) look like? Then, it will differ from JSON Schema provided one, and we could set it back then to `oneOf` if needed.

Btw, this change should be propagated to the `oneOf` in OpenAPI since OpenAPI schema is also a superset of JSON Schema and it includes the `$ref` property.


Additionally, I investigated a bit if we could get rid of those `if/then` and I believe we could just use `allOf` with all the possible cases as schemas right directly, without the `if/then`, including the case when the `schemaFormat` is just an unknown `string`.
But this, is out of the scope of this issue I believe. Just for the record.

smoya (changes requested):
Bundler looks ok, but the change in the definition shouldn't happen.

Please read my review comment! Feel free to reach me here or over Slack and happy to help!

smoya:
I need to re-review this PR since it has new changes since last time I reviewed.

I built up the following table to understand all use cases the `multiFormatSchema` should handle so we can use it for reviewing this PR. 

|     schemaFormat    | explicit schema | referenced ($ref) |         type         |
|:-------------------:|:---------------:|:-----------------:|:--------------------:|
|          -          |        -        |         -         |    AsyncAPI Schema   |
|          -          |        -        |         ✓         |  AsyncAPI Reference  |
|          -          |        ✓        |         -         |    AsyncAPI Schema   |
|       asyncapi      |        ✓        |         -         |    AsyncAPI Schema   |
|       asyncapi      |        -        |         ✓         |  AsyncAPI Reference  |
|      jsonschema     |        ✓        |         -         | JSON Schema Draft 07 |
|      jsonschema     |        -        |         ✓         |  AsyncAPI Reference  |
|       openapi       |        ✓        |         -         |  OpenAPI Schema 3.0  |
|       openapi       |        -        |         ✓         |  AsyncAPI Reference  |
|         avro        |        ✓        |         -         |    Avro Schema v1    |
|         avro        |        -        |         ✓         |  AsyncAPI Reference  |
| Non-matching string |        ✓        |         -         |        Object        |
| Non-matching string |        -        |         ✓         |  AsyncAPI Reference  |

Anyway, please folks check If I'm missing something just in case (pinging random maintainers) cc @fmvilas @char0n @GreenRover and @Pakisan

Pakisan:
@smoya check out new version. Works good outside JetBrains IDEA(crappy bug is still there, with recognition of Avro type for auto-completion)

smoya:
@Pakisan with your latest changes, validation is broken. You still need to split version of `anySchema` and `multiFormatSchema`, otherwise, you will need to have one version of each of the elements in the table I wrote in https://github.com/asyncapi/spec-json-schemas/pull/495#issuecomment-2014750383 as separate element inside the `anyOf` (which btw would need to be a `oneOf`).

Please revert back that side, then ensure `multiFormatSchema` fulfills with the table I wrote.

Pakisan:
> @Pakisan with your latest changes, validation is broken.
> 
> In my comment in [#495 (comment)](https://github.com/asyncapi/spec-json-schemas/pull/495#discussion_r1528737741) I mentioned the quickest simplest solution, and commented that removing the ifs could be done additionally but could be in another PR. If you still want to do it in that way, then keep reading, otherwise go for the simplest solution and open a follow up issue (and work on it if you want).
> 
> You still need to split version of `anySchema` and `multiFormatSchema`, otherwise, you will need to have one version of each of the elements in the table I wrote in [#495 (comment)](https://github.com/asyncapi/spec-json-schemas/pull/495#issuecomment-2014750383) as separate element inside the `anyOf` (which btw would need to be a `oneOf`).
> 
> Please revert back that side, then ensure `multiFormatSchema` fulfills with the table I wrote.

I have refactored `anySchema`, `multiformatSchema`. I think, that's enough for this issue. It's better than nothing, but this is a temporal solution

I'm sure that it's time to extend our test cases, and review all of current Schemas. 

Main problem, after addition of new key word `schemaFormat`, is collisions between Schema, Json Schema, Avro Schema with weak constraints for additional properties. That's why I have added `additionalProperties: false` to Avro, for example

jonaslagoni:
@Pakisan would highly recommend splitting the bundler fix for bindings into a separate PR so it's not blocked by the continued discussion regarding multiformat schema object.

> Main problem, after addition of new key word schemaFormat, is collisions between Schema, Json Schema, Avro Schema with weak constraints for additional properties. That's why I have added additionalProperties: false to Avro, for example

There really should be no collision when the validation occurs based on the conditioning of the value in `schemaFormat` 🤔 `additionalProperties` should have no effect other then for the schema format in use.

Pakisan:
> @Pakisan would highly recommend splitting the bundler fix for bindings into a separate PR so it's not blocked by the continued discussion regarding multiformat schema object.
> 
> > Main problem, after addition of new key word schemaFormat, is collisions between Schema, Json Schema, Avro Schema with weak constraints for additional properties. That's why I have added additionalProperties: false to Avro, for example
> 
> There really should be no collision when the validation occurs based on the conditioning of the value in `schemaFormat` 🤔 `additionalProperties` should have no effect other then for the schema format in use.

```json
{
  "$id": "http://asyncapi.com/definitions/3.0.0/anySchema.json",
  "type": "object",
  "description": "An object representing either a Reference, a Schema or a Multi Format Schema",
  "properties": {
    "schemaFormat": {
      "type": "string",
      "description": "Supported Schema format"
    },
    "schema": {
      "type": "object",
      "description": "Schema definition"
    },
    "$ref": {
      "type": "string",
      "description": "Reference to schema"
    }
  },
  "oneOf": [
    {
      "description": "Because of $ref collision in Reference and AsyncAPI Schema(includes $ref from Json Schema)",
      "not": {"required": ["schemaFormat", "schema"]},
    },
    {
      "type": "object",
      "required": ["schemaFormat", "schema"],
      "not": {"required": ["$ref"]},
      "minProperties": 2,
      "maxProperties": 2,
      "$ref": "http://asyncapi.com/definitions/3.0.0/multiFormatSchema.json"
    }
  ]
}
```

Here is simplified restriction to tell validator which schema to use. Our tests ok, because AJV is not in strict mode, but other validators, for example this will tell you that this two variants are satisfied:

Screenshot 2024-03-27 at 15 18 58

It's because it checks:
- `multiFormatSchema` requires this two fields - OK // expected
- `schema` allows additional properties - OK // not expected

and as result, it tells that `oneOf` is not satisfied

jonaslagoni:
@Pakisan I took a look at the old schemas and tested it locally which seems to work fine with just two changes.

1. remove disabling of `additionalProperties`: https://github.com/asyncapi/spec-json-schemas/blob/bee34d8569a18ee2117b4f2b0529a4e883e73638/definitions/3.0.0/multiFormatSchema.json#L13
2. remove the property `schema`: https://github.com/asyncapi/spec-json-schemas/blob/bee34d8569a18ee2117b4f2b0529a4e883e73638/definitions/3.0.0/multiFormatSchema.json#L53

<details>
  <summary>Full schema here:</summary>
  
```json
{
  "description": "The Multi Format Schema Object represents a schema definition. It differs from the Schema Object in that it supports multiple schema formats or languages (e.g., JSON Schema, Avro, etc.).",
  "if": {
    "not": {
      "type": "object"
    }
  },
  "then": {
  },
  "else": {
    "type": "object",
    "patternProperties": {
      "^x-[\\w\\d\\.\\x2d_]+$": {
        "$ref": "http://asyncapi.com/definitions/3.0.0/specificationExtension.json"
      }
    },
    "properties": {
      "schemaFormat": {
        "description": "A string containing the name of the schema format that is used to define the information. If schemaFormat is missing, it MUST default to application/vnd.aai.asyncapi+json;version={{asyncapi}} where {{asyncapi}} matches the AsyncAPI Version String. In such a case, this would make the Multi Format Schema Object equivalent to the Schema Object. When using Reference Object within the schema, the schemaFormat of the resource being referenced MUST match the schemaFormat of the schema that contains the initial reference. For example, if you reference Avro schema, then schemaFormat of referencing resource and the resource being reference MUST match.",
        "anyOf": [
          {
            "type": "string"
          },
          {
            "description": "All the schema formats tooling MUST support",
            "enum": [
              "application/schema+json;version=draft-07",
              "application/schema+yaml;version=draft-07",

              "application/vnd.aai.asyncapi;version=3.0.0",
              "application/vnd.aai.asyncapi+json;version=3.0.0",
              "application/vnd.aai.asyncapi+yaml;version=3.0.0"
            ]
          },
          {
            "description": "All the schema formats tools are RECOMMENDED to support",
            "enum": [
              "application/vnd.oai.openapi;version=3.0.0", 
              "application/vnd.oai.openapi+json;version=3.0.0", 
              "application/vnd.oai.openapi+yaml;version=3.0.0",
  
              "application/vnd.apache.avro;version=1.9.0",
              "application/vnd.apache.avro+json;version=1.9.0",
              "application/vnd.apache.avro+yaml;version=1.9.0",
              
              "application/raml+yaml;version=1.0"
            ]
          }
        ]
      }
    },
    "allOf": [
      {
        "if": {
          "not": {
            "description": "If no schemaFormat has been defined, default to schema or reference",
            "required": [
              "schemaFormat"
            ]
          }
        },
        "then": {
          "properties": {
            "schema": {
              "oneOf": [
                {
                },
                {
                }
              ]
            }
          }
        }
      },
      {
        "if": {
          "description": "If schemaFormat has been defined check if it's one of the AsyncAPI Schema Object formats",
          "required": [
            "schemaFormat"
          ],
          "properties": {
            "schemaFormat": {
              "enum": [
                "application/vnd.aai.asyncapi;version=2.0.0",
                "application/vnd.aai.asyncapi+json;version=2.0.0",
                "application/vnd.aai.asyncapi+yaml;version=2.0.0",
                "application/vnd.aai.asyncapi;version=2.1.0",
                "application/vnd.aai.asyncapi+json;version=2.1.0",
                "application/vnd.aai.asyncapi+yaml;version=2.1.0",
                "application/vnd.aai.asyncapi;version=2.2.0",
                "application/vnd.aai.asyncapi+json;version=2.2.0",
                "application/vnd.aai.asyncapi+yaml;version=2.2.0",
                "application/vnd.aai.asyncapi;version=2.3.0",
                "application/vnd.aai.asyncapi+json;version=2.3.0",
                "application/vnd.aai.asyncapi+yaml;version=2.3.0",
                "application/vnd.aai.asyncapi;version=2.4.0",
                "application/vnd.aai.asyncapi+json;version=2.4.0",
                "application/vnd.aai.asyncapi+yaml;version=2.4.0",
                "application/vnd.aai.asyncapi;version=2.5.0",
                "application/vnd.aai.asyncapi+json;version=2.5.0",
                "application/vnd.aai.asyncapi+yaml;version=2.5.0",
                "application/vnd.aai.asyncapi;version=2.6.0",
                "application/vnd.aai.asyncapi+json;version=2.6.0",
                "application/vnd.aai.asyncapi+yaml;version=2.6.0",
                "application/vnd.aai.asyncapi;version=3.0.0",
                "application/vnd.aai.asyncapi+json;version=3.0.0",
                "application/vnd.aai.asyncapi+yaml;version=3.0.0"
              ]
            }
          }
        },
        "then": {
          "properties": {
            "schema": {
              "oneOf": [
                {
                },
                {
                }
              ]
            }
          }
        }
      },
      {
        "if": {
          "required": [
            "schemaFormat"
          ],
          "properties": {
            "schemaFormat": {
              "enum": [
                "application/schema+json;version=draft-07",
                "application/schema+yaml;version=draft-07"
              ]
            }
          }
        },
        "then": {
          "properties": {
            "schema": {
              "oneOf": [
                {
                },
                {
                  "$ref": "http://json-schema.org/draft-07/schema"
                }
              ]
            }
          }
        }
      },
      {
        "if": {
          "required": [
            "schemaFormat"
          ],
          "properties": {
            "schemaFormat": {
              "enum": [
                "application/vnd.oai.openapi;version=3.0.0",
                "application/vnd.oai.openapi+json;version=3.0.0",
                "application/vnd.oai.openapi+yaml;version=3.0.0"
              ]
            }
          }
        },
        "then": {
          "properties": {
            "schema": {
              "oneOf": [
                {
                },
                {
                  "$ref": "http://asyncapi.com/definitions/3.0.0/openapiSchema_3_0.json"
                }
              ]
            }
          }
        }
      },
      {
        "if": {
          "required": [
            "schemaFormat"
          ],
          "properties": {
            "schemaFormat": {
              "enum": [
                "application/vnd.apache.avro;version=1.9.0",
                "application/vnd.apache.avro+json;version=1.9.0",
                "application/vnd.apache.avro+yaml;version=1.9.0"
              ]
            }
          }
        },
        "then": {
          "properties": {
            "schema": {
              "oneOf": [
                {
                },
                {
                  "$ref": "http://asyncapi.com/definitions/3.0.0/avroSchema_v1.json"
                }
              ]
            }
          }
        }
      }
    ]
  },
  "$id": "http://asyncapi.com/definitions/3.0.0/multiFormatSchema.json"
}
```
</details>


I tested it locally with Avro and OpenAPI references and it seems to have solved the problem without having to do major refactorings. Mind giving it a test?

jonaslagoni:
Technically we could add the following to `anySchema` (https://github.com/asyncapi/spec-json-schemas/blob/bee34d8569a18ee2117b4f2b0529a4e883e73638/definitions/3.0.0/anySchema.json#L10):

```
  "else": {
    "oneOf": [
      {
      },
      {
      }
    ]
  },
```

However, because `http://asyncapi.com/definitions/3.0.0/schema.json` references `http://json-schema.org/draft-07/schema` that already contains `$ref` property, it's not really needed.

Pakisan:
@jonaslagoni add from my revision(a4c5ef07) all test schemas

jonaslagoni:
@Pakisan testing with them now

jonaslagoni:
@Pakisan all seem to pass with the changes in https://github.com/asyncapi/spec-json-schemas/pull/495#issuecomment-2022561447

I can run `npm run test` with no problem and manually tested one of them.

definitions/3.0.0/components.json
```diff
@@ -13,14 +13,7 @@
...
         "^[\\w\\d\\.\\-_]+$": {
-          "oneOf": [
-            {
-              "$ref": "http://asyncapi.com/definitions/3.0.0/Reference.json"
-            },
-            {
-              "$ref": "http://asyncapi.com/definitions/3.0.0/anySchema.json"
-            }
-          ]
+          "$ref": "http://asyncapi.com/definitions/3.0.0/anySchema.json"
```
jonaslagoni, line 16: I would definitely keep this as well 👍

Pakisan:
> @Pakisan all seem to pass with the changes in [#495 (comment)](https://github.com/asyncapi/spec-json-schemas/pull/495#issuecomment-2022561447)
> 
> I can run `npm run test` with no problem and manually tested one of them.

Yep. I found that related MR is fix, not replacement for current. Let me apply your changes and check them

Pakisan:
@jonaslagoni applied your changes

https://github.com/asyncapi/spec-json-schemas/pull/495#issuecomment-2022561447

definitions/3.0.0/anySchema.json
```diff
@@ -1,16 +1,35 @@
...
+    },
+    "schema": {
+      "type": "object",
+      "description": "Schema definition"
+    },
+    "$ref": {
+      "type": "string",
+      "description": "Reference to schema"
+    }
+  },
```
jonaslagoni, line 19: This change should no longer be needed, original file should work fine :v:

definitions/3.0.0/multiFormatSchema.json
```diff
@@ -1,5 +1,11 @@
 {
   "description": "The Multi Format Schema Object represents a schema definition. It differs from the Schema Object in that it supports multiple schema formats or languages (e.g., JSON Schema, Avro, etc.).",
+  "type": "object",
+  "patternProperties": {
+    "^x-[\\w\\d\\.\\x2d_]+$": {
+      "$ref": "http://asyncapi.com/definitions/3.0.0/specificationExtension.json"
+    }
+  },
   "if": {
```
jonaslagoni, line 9: If you want, feel free to remove this if/then as https://github.com/asyncapi/spec-json-schemas/pull/507 does :v:

common/avroSchema_v1.json
```diff
@@ -70,7 +70,8 @@
       },
       "required": [
         "type"
-      ]
+      ],
+      "additionalProperties": false
```
jonaslagoni, line 74: I dont think these changes with `additionalProperties` is necessary, and on the other side, are you not allowed to have additional properties in Avro? 🤔

jonaslagoni (changes requested):

common/avroSchema_v1.json
Pakisan: As I remember Avro is allows only defined attributes. For example Record - https://avro.apache.org/docs/1.11.1/specification/#schema-record

definitions/3.0.0/anySchema.json
Pakisan: I removed properties, thanks for suggestion)

definitions/3.0.0/anySchema.json
```diff
@@ -1,16 +1,21 @@
...
+    },
+    {
+      "type": "object",
+      "required": ["schemaFormat", "schema"],
+      "not": {"required": ["$ref"]},
+      "minProperties": 2,
+      "maxProperties": 2,
+      "$ref": "http://asyncapi.com/definitions/3.0.0/multiFormatSchema.json"
+    }
+  ]
```
jonaslagoni, line 20: Meant the entire thing @Pakisan, `anySchema` does not need any changes to work 🙂

jonaslagoni (changes requested):

common/avroSchema_v1.json
jonaslagoni: Seems like additional properties is allowed on the different levels of Avro: https://asyncapi.slack.com/archives/C0230UAM6R3/p1712059889567579?thread_ts=1711713980.293369&cid=C0230UAM6R3

Pakisan: I removed restriction

jonaslagoni (approved):
👍

smoya (approved):
LGTM! 🚀🌔

Thanks for all the time invested in solving this issue @Pakisan. We all had to do a very focused review on this but finally got sorted out 🙌 

Thanks @jonaslagoni as well for your second review.

PR #506, opened 2024-03-24
feat: add extensions support

sambhavgupta0705:
This PR is part of asyncapi mentorship program
related issue: https://github.com/asyncapi/spec-json-schemas/issues/280

The previous was targeted to `next-major-spec` branch
All the changes have been copied and migrated to master branch

### DIFF
--- a/tools/bundler/index.js
+++ b/tools/bundler/index.js
@@ -1,8 +1,8 @@
 const path = require('path');
 const fs = require('fs');
 const traverse = require('json-schema-traverse');
-const { url } = require('inspector');
 const definitionsDirectory = path.resolve(__dirname, '../../definitions');
+const commonSchemasDirectory = path.resolve(__dirname, '../../common');
 const bindingsDirectory = path.resolve(__dirname, '../../bindings');
 const extensionsDirectory = path.resolve(__dirname, '../../extensions');
 const outputDirectory = path.resolve(__dirname, '../../schemas');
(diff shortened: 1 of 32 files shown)


sambhavgupta0705:
cc: @derberg  @smoya

sambhavgupta0705:
@derberg the previous pr was targetting to `next-spec-major` and this is targetting `master`.
I just copied the changes directly in this and now there are some deletions in https://github.com/asyncapi/spec-json-schemas/pull/506/files#diff-6deb33116f8f13032612ff1f8a6305914860872230d74fa630e77e3a277be4a2 for which I would like to know what should I do.

- should I only add extension related changes
- or should let this file as it is

derberg:
@sambhavgupta0705 sorry can you point to some example line of code and give more context, I'm not sure I follow

tools/bundler/index.js
```diff
@@ -1,27 +1,32 @@
 const path = require('path');
 const fs = require('fs');
 const traverse = require('json-schema-traverse');
+const { url } = require('inspector');
```
sambhavgupta0705, line 4: @derberg for example this
This wasn't in your pr but was added in `next-major-spec`

derberg: I don't see it being used in code

sambhavgupta0705: Yes 
When you raised a pr on my branch then it was there but after that it was removed from master

derberg (approved):
Approving and merging as it is a PR that I worked with @sambhavgupta0705 for quite some time already, https://github.com/asyncapi/spec-json-schemas/pull/443/ was just merged into a wring branch 😄

PR #501, opened 2024-03-14
fix: kafka bindings not using latest version

dalelane:
cf. spotted in a comment in
https://github.com/asyncapi/spec-json-schemas/pull/481

PR #508, opened 2024-03-31
fix: kafka bindings not using latest version

ctasada:
Added missing versions to https://github.com/asyncapi/spec-json-schemas/pull/501

Added fix as per https://github.com/asyncapi/spec-json-schemas/pull/501#issuecomment-2028640789


dalelane (approved):
thanks for fixing this - sorry I missed it!

PR #528, opened 2024-04-20
feat: add linkedin extension support

sambhavgupta0705:
this pr add essential schemas to add support for linkedin extension
https://github.com/asyncapi/extensions-catalog/issues/78
cc: @derberg

### DIFF
--- a/extensions/linkedin/0.1.0/schema.json
+++ b/extensions/linkedin/0.1.0/schema.json
@@ -2,7 +2,7 @@
     "type": "string",
     "patternProperties": {
         "x-linkedin":{
-            "$ref": "^https:\/\/(?:www\.)?linkedin\.com.*"
+            "$ref": "^https://(www\\.)?linkedin\\.com.*$"
         }
     },
     "description": "This extension allows you to provide the Linkedin profile URL of the account representing the team/company of the API.",
--- a/extensions/linkedin/0.1.0/schema.json
+++ b/extensions/linkedin/0.1.0/schema.json
@@ -1,8 +1,9 @@
 {
     "type": "string",
+    "format": "uri",
     "patternProperties": {
         "x-linkedin":{
-            "$ref": "^https://(www\\.)?linkedin\\.com*$"
+            "$ref": "^https://(www\\.)?linkedin\\.com.*$"
         }
     },
     "description": "This extension allows you to provide the Linkedin profile URL of the account representing the team/company of the API.",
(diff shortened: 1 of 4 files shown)


extensions/linkedin/0.1.0/schema.json
```diff
@@ -0,0 +1,10 @@
+{
+    "type": "string",
```
derberg, line 2: how about we also add `pattern` field, like `^https:\/\/(?:www\.)?linkedin\.com.*` that will validate provided link to make sure it starts with `https://www.linkedin.com/` or `https://linkedin.com/`

sambhavgupta0705: yes it will work

derberg: you used a wrong field

https://json-schema.org/understanding-json-schema/reference/regular_expressions

derberg (approved):
🚀

derberg:
Anyone else wanna have a look?

With this PR @sambhavgupta0705 basically validates that https://github.com/asyncapi/spec-json-schemas?tab=readme-ov-file#extensions instruction works. Also simplistic contribution flow to https://github.com/asyncapi/extensions-catalog/pull/144 works as well.

Once we merge this I think we can say we have super minor nice MVP of extension catalog:
- we can onboard new extensions in https://github.com/asyncapi/extensions-catalog
- and we can even onboard them in official json schema https://github.com/asyncapi/spec-json-schemas
- we also have automation to automatically render docs to new extensions in docs https://www.asyncapi.com/docs/reference/extensions/x
- we also figured the best way to have extensions support in react component https://github.com/asyncapi/asyncapi-react/issues/819 that is already in progress to be developed and contributed by folks from SAP. I asked them that in their contribution they could out of the box provide extension from `x-x` and `x-linkedin` extensions.

extensions/linkedin/0.1.0/schema.json
```diff
@@ -0,0 +1,11 @@
...
+    "pattern": "^https://(www\\.)?linkedin\\.com.*$",
```
smoya, line 3: IMHO Doesn't matter if users write `http` or rather `https`. Both are valid. It is linkedin who should decide if redirect will happen or not.

```suggestion
    "pattern": "^http(s)?://(www\\.)?linkedin\\.com.*$",
```

smoya (changes requested):
Simple change requested here https://github.com/asyncapi/spec-json-schemas/pull/528#discussion_r1577551366

The rest LGTM 👍 🚀

smoya (approved):
LGTM! 🚀🌔

Issue #364, opened 2023-03-16
Enable reusability of schemas between versions


jonaslagoni:
#### Reason/Context
There are specifically 2 schemas we wish to share among all versions.
- avroSchema_v1
- openapiSchema_3_0

To achieve this we need to do the following:
1. Move the common schemas to a common definition folder
2. Adapt the bundler (tools/bundler/index.js) to include these common definitions for all versions.

achiverram28:
I would like to work on this

devilkiller-ag:
Hi @jonaslagoni, is this issue still un-resolved? Also while going through the `tools/bundler/index.js` file, I found some repetitive spelling mistakes in code comments.

jonaslagoni:
Still a problem yea, feel free to grab it 🤙

devilkiller-ag:
Hi @jonaslagoni, Just to be clear, I need to first move these files from `definitions/v3.0.0` to a new folder `definitions/common` and then update the `loadDefinitions` function in `tools/bundler/index.js` accordingly?

AnimeshKumar923:
what's the status of this one? @jonaslagoni @devilkiller-ag

devilkiller-ag:
Hi @AnimeshKumar923 I am sorry, I forgot this one. I will complete this in the upcoming week as I am currently on vacation.

Shahbaz898414:
Hi @jonaslagoni, is this issue still unresolved?

jonaslagoni:
@Shahbaz898414 its being worked on yes

PR #533, opened 2024-05-13
feat!: remove duplicate schemas

jonaslagoni:
**Description**
Even though we enabled the reusability of schemas in https://github.com/asyncapi/spec-json-schemas/issues/364, the documents were never adapted, leaving a lot of unnecessary documents. This removes those and uses the common definitions.

This means there will no longer be anything called `http://asyncapi.com/definitions/x.x.x/avroSchema_v1.json` or `http://asyncapi.com/definitions/x.x.x/openapiSchema_3_0.json` but only `http://asyncapi.com/common/avroSchema_v1.json` and `http://asyncapi.com/common/openapiSchema_3_0.json`

**Related issue(s)**
Should have been fixed as part of https://github.com/asyncapi/spec-json-schemas/issues/364


jonaslagoni:
I see the bundler has some custom code for common, have to redefine that.

jonaslagoni:
Think we have to move this over to a `next` release branch

Pakisan (approved):
Looks valid for me

Pakisan (approved):

jonaslagoni:
@dalelane @lbroudoux because of the Kafka binding change Github need one of you to take a look when you can 🤙

smoya (approved):
LGTM! 🚀🌔

smoya (changes requested):
@jonaslagoni Not a good deal to have the need of serving those common schemas under the path `common` on the root of `http://asyncapi.com`. Not only because the need of adding an extra redirection but also because it is requiring a new path `/common` that is not so intuitive in terms of naming (what is common in relation to the AsyncAPI website?).

WDYT about `http://asyncapi.com/schemas/common`? This will require a move of the `common` dir under `schemas` directory.

jonaslagoni:
> @jonaslagoni Not a good deal to have the need of serving those common schemas under the path `common` on the root of `http://asyncapi.com`. Not only because the need of adding an extra redirection but also because it is requiring a new path `/common` that is not so intuitive in terms of naming (what is common in relation to the AsyncAPI website?).
> 
> WDYT about `http://asyncapi.com/schemas/common`? This will require a move of the `common` dir under `schemas` directory.

Schemas hold the bundled versions, and think we should keep it clean like that 🤔 

What about under `definitions`?

smoya:
> What about under `definitions`?

Fine with it.

Pakisan:
@jonaslagoni @smoya maybe repeat structure from jasyncapi?

Current tests, which I'm writing now, in other branch, were grouped by next logic

- models/[v2|v3]
- schemas/[openapi|asyncapi|avro|json]
- /schemas/asyncapi/security schemas/[v2|v3]/{security implementation}

They could be unified for all versions, but in v3 version one property was renamed for OAuth2 flow 😣

jonaslagoni:
> @jonaslagoni @smoya maybe repeat structure from jasyncapi?
> 
> Current tests, which I'm writing now, in other branch, were grouped by next logic
> 
>     * models/[v2|v3]
> 
>     * schemas/[openapi|asyncapi|avro|json]
> 
>     * /schemas/asyncapi/security schemas/[v2|v3]/{security implementation}
> 
> 
> They could be unified for all versions, but in v3 version one property was renamed for OAuth2 flow 😣

@Pakisan mind opening an issue for it? 🙂 It will for sure be lost in this PR 😅

dalelane (approved):

smoya (approved):
LGTM! 🚀🌔

PR #534, opened 2024-05-13
feat!: upgrade bundler 

jonaslagoni:
**Description**
We are using a very outdated version of the bundler, this PR upgrades it to the newest.


jonaslagoni:
Think we have to move this over to a `next` release branch

PR #540, opened 2024-05-22
test: 3.0.0 initial tests

### LINKED ISSUE #539: 3.0.0 initial tests
### Why do we need this improvement?

To be sure that our definitions are working like expected

### How will this change help?

see above

### Screenshots

_No response_

### How could it be implemented/designed?

- [x] APIKeyHTTPSecurityScheme.json
- [x] BearerHTTPSecurityScheme.json
- [x] HTTPSecurityScheme.json
- [x] NonBearerHTTPSecurityScheme.json
- [x] Reference.json
- [x] ReferenceObject.json
- [x] SaslGssapiSecurityScheme.json
- [x] SaslPlainSecurityScheme.json
- [x] SaslScramSecurityScheme.json
- [x] SaslSecurityScheme.json
- [x] SecurityScheme.json
- [x] X509.json
- [x] anySchema.json
- [x] apiKey.json
- [x] asymmetricEncryption.json
- [x] asyncapi.json
- [x] channel.json
- [x] channelBindingsObject.json
- [x] channelMessages.json
- [x] channels.json
- [x] components.json
- [x] contact.json
- [x] correlationId.json
- [x] externalDocs.json
- [x] info.json
- [x] infoExtensions.json
- [x] license.json
- [x] messageBindingsObject.json
- [x] messageExampleObject.json
- [x] messageObject.json
- [x] messageTrait.json
- [x] messages.json
- [x] multiFormatSchema.json
- [x] oauth2Flow.json
- [x] oauth2Flows.json
- [x] oauth2Scopes.json
- [x] openIdConnect.json
- [x] operation.json
- [x] operationBindingsObject.json
- [x] operationReply.json
- [x] operationReplyAddress.json
- [x] operationTrait.json
- [x] operations.json
- [x] parameter.json
- [x] parameters.json
- [x] schema.json
- [x] securityRequirements.json
- [x] server.json
- [x] serverBindingsObject.json
- [x] serverVariable.json
- [x] serverVariables.json
- [x] servers.json
- [x] specificationExtension.json
- [x] symmetricEncryption.json
- [x] tag.json
- [x] userPassword.json

### 🚧 Breaking changes

No

### 👀 Have you checked for similar open issues?

- [X] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue?

Yes I am willing to submit a PR!

Pakisan:
Changes:
- Vitest instead of Mocha - because I want work correctly with relative paths, and don't be afraid to move tests
- schemas base tests
- bindings tests
- common structure for tests - single entry point. Tests only provide test data, to work with

All ignored tests, with reasons are described here - https://github.com/asyncapi/spec-json-schemas/discussions/546

closes #539


test/definitions/3.0.0/models/channel/channel/channel.js
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  it('extended', () => {
+    const info = JSON.parse(fs.readFileSync(`${__dirname}/extended.json`, 'utf-8'));
+    const validationResult = validator(info);
+
+    assert(validationResult === true, `${jsonSchemaName} can be extended`);
+  });
+
+  it('wrongly extended', () => {
```
jonaslagoni, line 57: Why use such a huge file for something so simple as testing `wrong` property 🤨

test/ajv-schemes.js
```diff
@@ -0,0 +1,141 @@
+module.exports = schemesV3_0_0;
```
jonaslagoni, line 1: I would take advantage of the existing code in the bundler that loads the definitions, bindings and extensions: https://github.com/asyncapi/spec-json-schemas/blob/8ab1e6689c7cf1a3e8bf56d30dfb7f787c642a3d/tools/bundler/index.js#L107

This implementation is gonna be annoying long term 😅

jonaslagoni (changes requested):
@Pakisan do you mind turning on all tests? Its completely fine if they fail as either it's the example that are incorrect or the JSON schema file

Pakisan:
> @Pakisan do you mind turning on all tests? Its completely fine if they fail as either it's the example that are incorrect or the JSON schema file

If failed tests didn't block release or merges, I'm ok with it

test/definitions/3.0.0/models/channel/channel/channel.js
Pakisan: You mean json file to test? It was copy paste. I'll shortener them

jonaslagoni:
> If failed tests didn't block release or merges, I'm ok with it

It kinda will, but lets see whether it's the test that are wrong or the actual schemas first 🙂

Pakisan:
> > If failed tests didn't block release or merges, I'm ok with it
> 
> Lets see whether it's the test that are wrong or the actual schemas first 🙂

Let's merge it and solve skipped cases step by step

For example, Server example is not valid because `serverVariable.examples` is array, that's why:

Invalid
```json
{
      "description" : "This value is assigned by the service provider, in this example `gigantic-server.com`",
      "examples" : null,
      "enum" : null,
      "default" : "demo"
}
```

Valid
```json
{
  "description": "This value is assigned by the service provider, in this example `gigantic-server.com`",
  "default": "demo"
}
```

Because schema MUST be fixed, for `examples`, for example:

From this:
```json
{
  "examples": {
    "type": "array",
    "description": "An array of examples of the server variable.",
    "items": {
      "type": "string"
    }
  }
}
```

To this:
```json
{
  "examples": {
    "type": ["null", "array"],
    "description": "An array of examples of the server variable.",
    "items": {
      "type": "string"
    }
  }
}
```

test/ajv-schemes.js
Pakisan: Can you show valid example?

jonaslagoni:
> > > If failed tests didn't block release or merges, I'm ok with it
> > 
> > 
> > Lets see whether it's the test that are wrong or the actual schemas first 🙂
> 
> Let's merge it and solve skipped cases step by step
> 
> For example, Server example is not valid because `serverVariable.examples` is array, that's why:
> 
> Invalid
> 
> ```json
> {
>   "description": "This value is assigned by the service provider, in this example `gigantic-server.com`",
>   "examples": null,
>   "enum": null,
>   "default": "demo"
> }
> ```
> 
> Valid
> 
> ```json
> {
>   "description": "This value is assigned by the service provider, in this example `gigantic-server.com`",
>   "default": "demo"
> }
> ```
> 
> Because schema MUST be fixed, for `examples`, for example:
> 
> From this:
> 
> ```json
> {
>   "examples": {
>     "type": "array",
>     "description": "An array of examples of the server variable.",
>     "items": {
>       "type": "string"
>     }
>   }
> }
> ```
> 
> To this:
> 
> ```json
> {
>   "examples": {
>     "type": ["null", "array"],
>     "description": "An array of examples of the server variable.",
>     "items": {
>       "type": "string"
>     }
>   }
> }
> ```

But setting examples to `null` is not allowed in the spec? https://www.asyncapi.com/docs/reference/specification/v3.0.0#serverVariableObject

Only extensions and `address` (on channel) are allowed to be null.

jonaslagoni (commented):
So far I can see the following skipped tests (that need to be addressed):
- AMQP 0.2.0 channel binding, 3 skipped test
- AMQP 0.3.0 channel binding, 3 skipped test
- SNS 0.1.0 operation binding, all tests skipped
- Solace 0.2 operation binding, 2 test skipped
- Solace 0.3 operation binding, 2 test skipped
- Solace 0.4 operation binding, 2 test skipped
- SQS 0.2 channel binding, 2 test skipped

Did I miss any @Pakisan?

Should we remove any old binding version tests? I.e. you both have test for AMQP 0.2 and 0.3, would it not be enough with 0.3 in this initial PR?

jonaslagoni:
>  Current implementation tells that null is not allowed, which is not correct

@Pakisan do you mean current implementation of the JSON Schema files or?

If yes, then no, `null` are not allowed, there is a difference between a property being set and a `null` value 🙂

Pakisan:
> > Current implementation tells that null is not allowed, which is not correct
> 
> @Pakisan do you mean current implementation of the JSON Schema files or?
> 
> If yes, then no, `null` are not allowed, there is a difference between a property being set and a `null` value 🙂

I don't sure that I got your point

Property with `null` is critical for [Json Merge Patch(RFC 7386)](https://datatracker.ietf.org/doc/html/rfc7386)

> Null values in the merge patch are given special meaning to indicate the removal of existing values in the target.

In other cases it's ok to write or ignore property with `null` value to show that it doesn't contain any meaningful value

On deserialization you will interpret `null` value or missing property as scenario where it will be `null` or default value

jonaslagoni:
> I don't sure that I got your point
> Property with null is critical for [Json Merge Patch(RFC 7386)](https://datatracker.ietf.org/doc/html/rfc7386)

In JSON there is a difference between allowing `{examples: null}` and not setting the property at all `{}`. `{examples: null}` according to the specification, is NOT allowed, so we cannot change that in the schema to be allowed, as you propose 🙂

> In other cases it's ok to write or ignore property with null value to show that it doesn't contain any meaningful value

> On deserialization you will interpret null value or missing property as scenario where it will be null or default value 

Be careful about bringing Java-specific implementation details into a generic JSON schema validation library. I.e. yes in Java there is no difference between a `null` value and optional property, i.e. when you serialize an object it can come out as `{examples: null}`. However, the specification does not care about implementation details, only the correct JSON values.

I.e. `{examples: null}` is not allowed.

If you use something like Jackson for the serialization process, I have tried with something like this: https://www.baeldung.com/jackson-optional before to allow both `null` values and optional properties.

> Property with null is critical for [Json Merge Patch(RFC 7386)](https://datatracker.ietf.org/doc/html/rfc7386)

That's a good point! However, it is beyond this discussion as that is a specification discussion on how to handle those cases where you want to remove existing properties when used in conjunction with traits 🤔

As of right now, only places `null` JSON values are allowed are extensions and `address` 🙂

Pakisan:
Here is `ServerVariable` class

```java
@Data
@Builder
@NoArgsConstructor
@AllArgsConstructor
@EqualsAndHashCode(callSuper = true)
public class ServerVariable extends ExtendableObject {

    /**
     * An enumeration of string values to be used if the substitution options are from a limited set.
     */
    @Nullable
    @JsonProperty(value = "enum")
    private List<String> enumValues;

    /**
     * The default value to use for substitution, and to send, if an alternate value is not supplied.
     */
    @Nullable
    @JsonProperty("default")
    private String defaultValue;

    /**
     * An optional description for the server variable. <a href="https://spec.commonmark.org/">CommonMark syntax</a> MAY be used for rich text representation.
     */
    @Nullable
    private String description;

    /**
     * An array of examples of the server variable.
     */
    @Nullable
    private List<String> examples;

}
```

I don't use `Optional<T>` on class fields, cause they were designed for other purposes (args, for example), so I don't propose to implement language-specific mechanisms

In `ServerVariable` Json Schema here are not `required`, that's why me, generators, other folks can interpret it like allowance to avoid property or set it to `null`

> In JSON there is a difference between allowing {examples: null} and not setting the property at all {}. {examples: null} according to the specification, is NOT allowed, so we cannot change that in the schema to be allowed, as you propose 🙂

Looks like you are talking about `ECMAScript Language Specification`, but Json ([ECMA-404 The JSON Data Interchange Standard](https://www.ecma-international.org/publications-and-standards/standards/ecma-404/)) says that:

>  However, it does not attempt to impose ECMAScript’s internal data representations on other programming languages. Instead, it shares a small subset of ECMAScript’s syntax with all other programming languages.

This interpretation of `null` as `{}` looks more like ECMA story

For Java and Go, for example:

1. We can't interpret `{}` as valid `null`, `nil` value on deserialization, out of the box
2. `null`(Java) or `nil`(Go) will be mapped to `null` on serialization

Go: https://pkg.go.dev/encoding/json
> Pointer values encode as the value pointed to. A nil pointer encodes as the null JSON value.

Java: User can choose interpret `null` as no field in object, of field with `null` value

That's why I propose to move out this point and  discuss it with other maintainers of our Json Schemas and call other folks from DX and Building Blocks working groups

@asyncapi/developer_experience_wg @asyncapi/essential_building_blocks_wg

jonaslagoni:
The only question at hand is, what is a valid AsyncAPI document, it does not matter in what language, it's an indisputable truth, no matter the implementation 🙂 And even though we want the JSON Schema files to be that source of truth, it will be whatever is written in https://www.asyncapi.com/docs/reference/specification/v3.0.

```yaml
asyncapi: 3.0.0
info:
  title: example with null and optional
  version: 1.0.0
servers:
  test:
    host: test
    protocol: http
    variables:
      test:
        examples: null
```

According to https://www.asyncapi.com/docs/reference/specification/v3.0.0#serverVariableObject (right this second) will never be considered valid.

```yaml
asyncapi: 3.0.0
info:
  title: example with null and optional
  version: 1.0.0
servers:
  test:
    host: test
    protocol: http
    variables:
      test: {}
```

However is valid, because `examples` are an optional property.

Therefore it's not a question of whether a language outputs it like the invalid example, it will still be considered invalid and other tools are not expected to say it is. 

If you want to enable `null` values i.e. the current invalid example above, it needs to be a feature request to https://github.com/asyncapi/spec 😄

Pakisan:
@jonaslagoni agree. Will create proposal to add null and create discussion to discuss current state of Schemas

What I need to improve, to merge tests as is and fix them step by step, binding each case with topic in discussion?

VisualBean:
From reading the spec, the example given with 'null' as a value is wrong.
Spec specifically states 
> examples | [string] | An array of examples of the server variable.

So not serializing it seems correct as a null array is not an example.
An array with a value that is null however, should be correct.

test/bindings/amqp/0.2.0/message/extended.json
```diff
@@ -0,0 +1,10 @@
+{
+    "contentEncoding": "gzip",
+    "messageType": "user.signup",
+    "bindingVersion": "0.2.0",
+    "x-number": 0,
+    "x-string": "",
+    "x-object": {
+        "property": {}
```
VisualBean, line 8: are empty properties supposed to be serialized?
From a specification perspective, it doesnt make much sense, as this isnt actually communicating much in terms of that property.

smoya:
I would appreciate if you summarize what should I review from this PR. I see **tons** of kinda repeated tests and I just want to understand the expectation.

Pakisan:
@jonaslagoni @smoya

Basic test structure is next:
- schema:
  - empty
  - example
  - without required fields
  - only required fields
  - extended
  - wrongly extended

Let's merge it. I'll extend tests under this issue #551

test/ajv-schemes.js
Pakisan: Moved up to`TestHelper` class

test/definitions/3.0.0/models/channel/channel/channel.js
Pakisan: Simplified

test/ajv-schemes.js
jonaslagoni: I can optimize it after

jonaslagoni (commented):
LGTM :+1:

Pakisan:
> LGTM :+1:

Can't find review, which is blocking merge, when approves appear

jonaslagoni:
@fmvilas @derberg @dalelane @smoya @char0n @GreenRover any of you want to have a quick look before merging it?

char0n:
Hi everybody,

One thing I've noticed are fixture files containing empty spaces. Example: `./without required properties.json`

Wouldn't kebab-case or snake-case be more appropriate here?

I would also probably inquire why `mocha` was replaced with `vitest`? Was there any actual test case `mocha` couldn't handle?

There is also an unanswered comment from @smoya - https://github.com/asyncapi/spec-json-schemas/pull/540#issuecomment-2161251446. Answering it would help me as well. 


Thanks for this huge work!

Pakisan:
> Hi everybody,
> 
> One thing I've noticed are fixture files containing empty spaces. Example: `./without required properties.json`
> 
> Wouldn't kebab-case or snake-case be more appropriate here?
> 
> I would also probably inquire why `mocha` was replaced with `vitest`? Was there any actual test case `mocha` couldn't handle?
> 
> There is also an unanswered comment from @smoya - [#540 (comment)](https://github.com/asyncapi/spec-json-schemas/pull/540#issuecomment-2161251446). Answering it would help me as well.
> 
> Thanks for this huge work!

Hi!

I have summarized here - https://github.com/asyncapi/spec-json-schemas/pull/540#issuecomment-2199862506

New tests will look like this, so no files

```js
import {describe} from 'vitest';
import {
    JsonSchemaTestSuite,
    JsonSchemaTestSuiteConfig,
    JsonSchemaTestSuiteData
} from '@test/definitions/base-schema-test.mjs';

const example = {
    "name": "AsyncApi",
    "url": "https://www.asyncapi.com",
    "email": "java@asyncapi.com"
}

const onlyRequiredProperties = {}

const withoutRequiredProperties = {}

const extended = {
    "name": "AsyncAPI",
    "url": "https://www.asyncapi.com",
    "email": "java@asyncapi.com",
    "x-number": 0,
    "x-string": "",
    "x-object": {
        "property": {}
    }
}

const wronglyExtended = {
    "name": "AsyncAPI",
    "url": "https://www.asyncapi.com",
    "email": "java@asyncapi.com",
    "x-number": 0,
    "x-string": "",
    "x-object": {
        "property" : { }
    },
    "ext-number": 1
}

const jsonSchema = require('@definitions/3.0.0/contact.json');

const config = new JsonSchemaTestSuiteConfig(
  false,
  [],
  true,
  []
);

const data = new JsonSchemaTestSuiteData(
  jsonSchema,
  [example],
  onlyRequiredProperties,
  withoutRequiredProperties,
  extended,
  wronglyExtended
);

describe('Contact', () => {
    new JsonSchemaTestSuite(data, config).testSuite()
});
```

I'll inject basic test resources to class. More specified cases will be stored nearly, in JSON, in future

> I would also probably inquire why `mocha` was replaced with `vitest`? Was there any actual test case `mocha` couldn't handle?

Vitest and mocha are backward compatible. I switched to Vitest to introduce aliases like `@test/definitions/base-schema-test.mjs` instead of relative paths

aeworxet:
@asyncapi/bounty_team

Pakisan:
@fmvilas @derberg @dalelane @smoya @char0n @GreenRover @jonaslagoni

Check out new tests

tldr;

- 95 test files for `definitions` and `bindings` instead of 400+, previously in this MR
- common tests structure - empty, extended, wrongly extended, only required, without required, examples

dalelane:
There is a lot here, so I'll likely have more comments once I've had a proper look through. 

My initial reaction is mostly to the use of spaces in the file names for the new files - I realise this is very much a personal style/preference thing, so I'm sure it works fine as-is, but I would've preferred these to be hyphenated (such as the files in https://github.com/asyncapi/spec-json-schemas/tree/master/migrations and https://github.com/asyncapi/spec-json-schemas/tree/master/scripts ) or camel-cased (such as the files in https://github.com/asyncapi/spec-json-schemas/tree/master/definitions/3.0.0 )

I'll have a proper look and try and add some more substantive comments later

jonaslagoni (commented):
Just a word of advice, nobody likes reviewing a 95 file change PR, and some maintainers might even force you to split it up. Also if the new testing setup is unfavored by majority maintainers you could end up having to rewrite it... So small PRs are always preferred! 

Other then that, this looks goot to me 👍 

However CI does not like it: https://github.com/asyncapi/spec-json-schemas/actions/runs/10109288041/job/27956881619?pr=540

GreenRover:
I am out with review. This is to much to understand what is the target and how things belong to each other. 
A review would take multiple hours.

Pakisan:
folks, sorry for delay

I understand, that review 95 files may look too much, but it's ordinary tests with basic structure 

It's not feature implementation, but only tests, don't be afraid 😅 

Proposal of this PR is enable basic tests, and tune them on a go

Right now, several tests, were disabled until schema fixes

All ignored tests, with reasons are described here - https://github.com/asyncapi/spec-json-schemas/discussions/546

🚀 : @char0n @fmvilas @smoya @dalelane @derberg @jonaslagoni

aeworxet:
/ptal

> it's ordinary tests with basic structure

jonaslagoni (approved):
Think we waited long enough, thanks for the patience @Pakisan, but as others mentioned, it is indeed a huge PR. But as this is tests and there are no explicit objections, they do not have any effect on the library itself, let's move this forward 🙂 

> My initial reaction is mostly to the use of spaces in the file names for the new files - I realise this is very much a personal style/preference thing, so I'm sure it works fine as-is, but I would've preferred these to be hyphenated (such as the files in https://github.com/asyncapi/spec-json-schemas/tree/master/migrations and https://github.com/asyncapi/spec-json-schemas/tree/master/scripts ) or camel-cased (such as the files in https://github.com/asyncapi/spec-json-schemas/tree/master/definitions/3.0.0 )

Regarding the comment from @dalelane I concur, please make an issue after this is merged so we can rectify this, I have a script that can easily do it🙂

jonaslagoni:
@Pakisan is there any other followup issues that are needed 🙂?

Pakisan:
@jonaslagoni

Next steps:
- resolve conflicts https://github.com/asyncapi/spec-json-schemas/pull/549
- create task to migrate this tests to tck bundled with https://github.com/sourcemeta/jsonschema validator
- patch found issues with each schema in repo, from https://github.com/asyncapi/spec-json-schemas/discussions/546

PR #552, opened 2024-06-15
refactor: included caching in add-new-version.js.

Manush-2005:
**Description**

- This file had too many fs read and write operations.
- Included cache which reads data from all files and returns the writes final cache at the end of execution( addNewVersion function).
- ReadJsonFile - Reads File and puts it in cache. 
- WriteJsonFile - Required cache file is written back.


Manush-2005:
Love to hear why this PR is still open @sambhavgupta0705

PR #558, opened 2024-09-30
docs: adding optional version to the message object

IsmaelMartinez:
**Description**

Add optional version to the message object
Part of the https://github.com/asyncapi/spec/issues/1068 RFC 0

**Related issue(s)**

https://github.com/asyncapi/spec/issues/1068
https://github.com/asyncapi/spec/pull/1069

### DIFF
--- a/schemas/3.0.0-without-$id.json
+++ b/schemas/3.0.0-without-$id.json
@@ -2542,10 +2542,6 @@
                     "type": "string",
                     "description": "A brief summary of the message."
                 },
-                "version": {
-                    "type": "string",
-                    "description": "A version for the message."
-                },
                 "name": {
                     "type": "string",
                     "description": "Name of the message."
--- a/schemas/3.0.0-without-$id.json
+++ b/schemas/3.0.0-without-$id.json
@@ -2542,6 +2542,10 @@
                     "type": "string",
                     "description": "A brief summary of the message."
                 },
+                "version": {
+                    "type": "string",
+                    "description": "A version for the message."
+                },
                 "name": {
                     "type": "string",
                     "description": "Name of the message."
--- a/schemas/3.0.0.json
+++ b/schemas/3.0.0.json
@@ -2594,6 +2594,10 @@
                     "type": "string",
                     "description": "A brief summary of the message."
                 },
+                "version": {
+                    "type": "string",
+                    "description": "A version for the message."
+                },
                 "name": {
                     "type": "string",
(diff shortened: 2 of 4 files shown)


schemas/3.0.0-without-$id.json
```diff
@@ -2542,6 +2542,10 @@
                     "type": "string",
                     "description": "A brief summary of the message."
                 },
+                "version": {
+                    "type": "string",
+                    "description": "A version for the message."
```
fmvilas, line 2547: Slight suggestion, feel free to ignore:

```suggestion
                    "description": "Version of the message."
```

schemas/3.0.0.json
```diff
@@ -2594,6 +2594,10 @@
                     "type": "string",
                     "description": "A brief summary of the message."
                 },
+                "version": {
+                    "type": "string",
+                    "description": "A version for the message."
```
fmvilas, line 2599: Slight suggestion, feel free to ignore:

```suggestion
                    "description": "Version of the message."
```

fmvilas (commented):
Looks good to me. I'm not approving because, although I'm a maintainer of the repo, I've been completely disconnected from the work it's happening here and I'm not sure these are the files we should be touching 😅 Therefore, leaving it to others who are more engaged.

Pakisan:
@IsmaelMartinez looks good

Waiting for test to test new property

jonaslagoni (changes requested):
@IsmaelMartinez make sure to change the `definitions` not the `schemas` as they are automatically generated on each release 🙂 

You can find the message object here: https://github.com/asyncapi/spec-json-schemas/blob/master/definitions/3.0.0/messageObject.json

IsmaelMartinez:
Thanks for the review. I will see if I can make those changes later on. 1st time in here so apologies for silly mistakes ;)

schemas/3.0.0-without-$id.json
IsmaelMartinez: applying this to the definition, as I have wrongly done it in the schema. Learning

schemas/3.0.0.json
IsmaelMartinez: same as above, I will apply it to the definitions. Thanks for the review

IsmaelMartinez:
Hi all, I just made the changes. We got an conversation open on the RFC but hopefully we can reach a consensus in the near future.

PR #561, opened 2024-12-30
fix: fix type of payload in MessageExample

### LINKED ISSUE #560: [BUG] fix type of payload in MessageExample
### Describe the bug.

Wrong definition of `payload` property

https://github.com/asyncapi/spec-json-schemas/blob/2da0459d619de3e2d05ca62ca8006dafa3d311f1/definitions/3.0.0/messageExampleObject.json#L25-L31

### Expected behavior

`MessageExample.payload`  must be defined as `any` type

```json
{
    "headers": {
      "type":  "object",
      "description": "Example of the application headers. It can be of any type."
    },
    "payload": {
      "description": "Example of the message payload. It can be of any type."
    }
}
```

See:
- https://github.com/asyncapi/jasyncapi/pull/208
- https://github.com/asyncapi/website/issues/3530

### 👀 Have you checked for similar open issues?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue ?

Yes I am willing to submit a PR!

Pakisan:
Define `payload` as `any` type

Fixes #560

### DIFF
--- a/definitions/3.0.0/messageExampleObject.json
+++ b/definitions/3.0.0/messageExampleObject.json
@@ -24,7 +24,7 @@
     },
     "headers": {
       "type": "object",
-      "description": "Example of the application headers. MUST be a map of key-value pairs."
+      "description": "Example of the application headers. It MUST be a map of key-value pairs."
     },
     "payload": {
       "type": ["number", "string", "boolean", "object", "array", "null"],
--- a/definitions/3.0.0/messageExampleObject.json
+++ b/definitions/3.0.0/messageExampleObject.json
@@ -24,7 +24,7 @@
     },
     "headers": {
       "type": "object",
-      "description": "Example of the application headers. It can be of any type."
+      "description": "Example of the application headers. MUST be a map of key-value pairs."
     },
     "payload": {
       "type": ["number", "string", "boolean", "object", "array", "null"],
--- a/definitions/3.0.0/messageExampleObject.json
+++ b/definitions/3.0.0/messageExampleObject.json
@@ -23,7 +23,7 @@
       "description": "A brief summary of the message example."
     },
     "headers": {
-      "type": ["number", "string", "boolean", "object", "array", "null"],
+      "type": "object",
       "description": "Example of the application headers. It can be of any type."
     },
     "payload": {
--- a/definitions/3.0.0/messageExampleObject.json
+++ b/definitions/3.0.0/messageExampleObject.json
@@ -23,10 +23,11 @@
       "description": "A brief summary of the message example."
     },
     "headers": {
-      "type": "object",
+      "type": ["number", "string", "boolean", "object", "array", "null"],
       "description": "Example of the application headers. It can be of any type."
     },
     "payload": {
+      "type": ["number", "string", "boolean", "object", "array", "null"],
       "description": "Example of the message payload. It can be of any type."
     }
   },


Pakisan:
@derberg @fmvilas @jonaslagoni @smoya

definitions/3.0.0/messageExampleObject.json
```diff
@@ -23,7 +23,7 @@
       "description": "A brief summary of the message example."
     },
     "headers": {
-      "type": ["number", "string", "boolean", "object", "array", "null"],
+      "type": "object",
       "description": "Example of the application headers. It can be of any type."
```
fmvilas, line 27: Please change the description too 🙏

fmvilas (changes requested):

definitions/3.0.0/messageExampleObject.json
Pakisan: Done

definitions/3.0.0/messageExampleObject.json
```diff
@@ -24,9 +24,10 @@
     },
     "headers": {
       "type": "object",
-      "description": "Example of the application headers. It can be of any type."
+      "description": "Example of the application headers. MUST be a map of key-value pairs."
```
fmvilas, line 27:
```suggestion
      "description": "Example of the application headers. It MUST be a map of key-value pairs."
```

definitions/3.0.0/messageExampleObject.json
```diff
@@ -24,9 +24,10 @@
...
     },
     "payload": {
+      "type": ["number", "string", "boolean", "object", "array", "null"],
```
fmvilas, line 30: Why maintaining a list of types here? 🤔

fmvilas (changes requested):

definitions/3.0.0/messageExampleObject.json
Pakisan: In previous versions here was `any` type. I found in SO that's better way to simulate any type and don't broke validation is to enumerate each JSON Schema type

fmvilas: Why is it going to break validation? When `type` is not specified, it can be any type, right? If some tools don't get it right, shouldn't it be fixed in the tools instead?

fmvilas: It's either a string or an array of strings but it's not required, so we can decide not to put it there. Effectively, I think it means "any" but good idea pinging the JSON Schema folks as, in the end, they know best.

Pakisan: Anyway, I'm ready to remove `type` and patch our schema for JetBrains IDE and other validators, as I have already done, previously

My main concern is to try to bring changes back, here, to make our schema compatible with various IDEs and environments where AsyncAPI specification will be validated via our JSON Schemas

And be sure that this changes are synced with JSON Schema specification requirements

fmvilas: Understandable mate. It's a fine line between trying to be pragmatic and keeping the spec strict. I'm not against maintaining this list of types but code you don't write is code you don't have to maintain. Arguably, explicit is better than implicit. Let's see what others have to say 🤷‍♂️ I'm not going to block this PR because of this line.

derberg:
yo, how is this one going? we need it for https://github.com/asyncapi/spec/pull/1082 to be merged

fmvilas (approved):
Approving as the discussion is non-blocking for me.

PR #582, opened 2025-06-24
feat: add ros2 bindings

amparo-siemens:
**Description**

- Adds support for ROS 2 bindings to based on https://github.com/asyncapi/bindings/pull/270

### DIFF
--- a/bindings/ros2/0.1.0/operation.json
+++ b/bindings/ros2/0.1.0/operation.json
@@ -1,0 +1,75 @@
+{
+  "type": "object",
+  "description": "This object contains information about the operation representation in ROS 2.",
+  "additionalProperties": false,
+  "patternProperties": {
+    "^x-[\\w\\d\\.\\x2d_]+$": {
+      "$ref": "http://asyncapi.com/definitions/3.0.0/specificationExtension.json"
+    }
+  },
+  "required": [
+    "role",
+    "node"
+  ],
+  "properties": {
+    "role": {
+      "type": "string",
+      "description": "Specifies the ROS 2 type of the node for this operation.",
+      "enum": ["publisher", "action_client", "service_client", "subscriber", "action_server", "service_server"]
+    },
+    "node": {
+      "type": "string",
+      "description": "The name of the ROS 2 node that implements this operation."
+    },
+    "qosPolicies": {
+      "type": "object",
+      "properties": {
+        "reliability": {
+            "type": "string",
+            "description": "Specifies the communication guarantee model that determines whether message delivery confirmation between publisher and subscriber is required.",
+            "enum": ["best_effort", "realiable"]
+        },
+        "history": {
+            "type": "string",
+            "description": "Policy parameter that defines the maximum number of samples maintained in the middleware queue",
+            "enum": ["keep_last", "keep_all", "unknown"]
+        },
+        "durability": {
+            "type": "string",
+            "description": "Persistence specification that determines message availability for late-joining subscribers",
+            "enum": ["transient_local", "volatile"]
+        },
+        "lifespan": {
+            "type": "integer",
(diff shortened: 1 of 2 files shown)


Pakisan (approved):
lgtm

Pakisan:
> Is there anything pending here @Pakisan? 😄

Sorry, lost focus. Thought that everything is ok and will be merged automatically

Looking what's went wrong

Pakisan:
> > Is there anything pending here @Pakisan? 😄
> 
> Sorry, lost focus. Thought that everything is ok and will be merged automatically
> 
> Looking what's went wrong

@amparo-siemens MR was patched

Check it out and after that it will be ready to merge

fyi: @derberg @char0n @jonaslagoni

Pakisan (approved):
/lgtm

Pakisan:
@amparo-siemens merged

Welcome on board 🚀

derberg:
@Pakisan @jonaslagoni hey, as I see here, the PR adds schemas to v3 which was released 2y ago. We need v3.1 release for that. Look into the main issue in the specification repo: https://github.com/asyncapi/spec/pull/1109. This release should actually be reverted.

derberg:
But this is a completely new binding

example, server bindings: https://www.asyncapi.com/docs/reference/specification/v3.0.0#serverBindingsObject

- there are fixed fields, and as you know ROS2 is not there
- and under the list of fixed fields you see "This object MAY be extended with Specification Extensions"

@jonaslagoni well spec doesn't say explicitly which parts, if modified, should trigger release, but then you could say that if we add new field to for example [Message Object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#messageObject) then it also should not trigger a release of the spec 😃 

We do have a definition of what is a breaking change and what is not a breaking change: https://github.com/asyncapi/spec/blob/master/CONTRIBUTING.md#breaking-change-vs-non-breaking-change. `ROS2` related objects are new properties added to spec. It is a change to the spec, non breaking, so minor release is needed.

in the past every new binging was a minor release:
- https://github.com/asyncapi/spec/pull/666
- https://github.com/asyncapi/spec/pull/545
- https://github.com/asyncapi/spec/pull/537

derberg:
bindings are not mandatory like many other properties/fields in the spec - this has nothing to do with "release or not release".

> Does it mean that every new version of already existed binding will trigger new release?

no, bindings releases are independent, but first, binding must be added to the spec. AsyncAPI specification do not contain details of specific binding, so we are cool here.

in other words:
- we need new release of AsyncAPI specification when new binding is added, as it is a new official field added to the spec. But the object defined under this new field, it is managed in `bindings` repository independently. So first `RSO2` release must go with AsyncAPI spec release.
- Later, any changes to the binding, of course by following semver,  like new fields, or removal of fields is done independently from specification releases. So you can have AsyncAPI v3.1 that will support `RSO2` in version `0.0.1` and `0.0.2` for example.

jonaslagoni:
I guess it makes sense 😄

Pakisan:
@derberg understood

I'll rollback this release and create new version using  instructions in readme

Pakisan:
@derberg tagged you in MR

@amparo-siemens will be available from upstream channel in new upcoming version, but still available in my ["nightly"](https://asyncapi.pavelon.dev/bindings/ros2/) channel (portal where I store tuned versions of our schemas for other tools and vendors)

gramss:
Thank you @derberg for clarifying the details with the release.

I think there might be a few things that can grow into and at some point out of the `ROS 2` binding.
Let's get our full tooling (introspection and code generator) rolled out based on this and then see where it will grow into.

Already, I liked the involvement of the ROS community and would like to see it as a shared effort in order to have not a single (tunnel) vision for the future of this binding. Maybe @Achllle might want to join this meeting with @derberg ? @amparo-siemens and I will sure participate!

Enjoy the summer and see you in September!

